PR #3, opened 2020-09-18
Dashboard performance improvements

pkmitre:
Improve dashboard performance by

1. Sharing the primary VA data query across all display components that need it
2. Optimizing the VA data query itself

### DIFF
--- a/va_explorer/va_analytics/dash_apps/va_dashboard.py
+++ b/va_explorer/va_analytics/dash_apps/va_dashboard.py
@@ -140,12 +140,12 @@ GEOJSON = load_geojson_data(json_file=f"{JSON_DIR}/{JSON_FILE}")
 
 # ============ VA Data =================
 def load_va_data(geographic_levels=None):
-    print("--> LOAD_VA_DATA CALLED")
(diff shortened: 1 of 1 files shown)


va_explorer/va_analytics/dash_apps/va_dashboard.py
```diff
@@ -574,20 +573,70 @@ def get_metric_display_names(map_metrics):
     ],
 )
 
+# ============ VA data (loaded from database and shared across components) ========
+
+@app.callback(
+    Output(component_id="va_data", component_property="children"),
+    [
+        Input(component_id="timeframe", component_property="value"),
```
babraham33, line 581: Currently not using this. Should it be passed to load_va_data()?

babraham33 (approved):

va_explorer/va_analytics/dash_apps/va_dashboard.py
pkmitre: I included it because

1) I *think* callbacks need an input of some sort? Could be wrong about that
2) We may want to use that in the query at some point

PR #4, opened 2020-09-24
added logic to handle lowercase columns for va data ingest

babraham33:
Added logic to load_va_csv to map lowercase columns names to corresponding model fields. The function should now be able to read in both lowercase and uppercase columns

### DIFF
--- a/verbal_autopsy/management/commands/load_va_csv.py
+++ b/verbal_autopsy/management/commands/load_va_csv.py
@@ -21,6 +21,11 @@ class Command(BaseCommand):
 
         # Figure out the common field names across the CSV and our model
         model_field_names = pd.Index([f.name for f in VerbalAutopsy._meta.get_fields()])
+        
+        # But first, account for case differences in csv columns (i.e. ensure id10041 maps to Id10041)
+        fieldCaseMapper = {field.lower(): field for field in model_field_names} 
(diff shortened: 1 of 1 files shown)


verbal_autopsy/management/commands/load_va_csv.py
```diff
@@ -21,6 +21,12 @@ def handle(self, *args, **options):
 
         # Figure out the common field names across the CSV and our model
         model_field_names = pd.Index([f.name for f in VerbalAutopsy._meta.get_fields()])
+        
+        # But first, account for case differences in csv columns (i.e. ensure id10041 maps to Id10041)
+        fieldCaseMapper = {field.lower(): field for field in model_field_names} 
+        csv_data.rename(columns=lambda c: fieldCaseMapper.get(c, c), inplace=True)
```
pkmitre, line 27: This is an elegant approach, but it only handles the case of a file with fully lowercase field names. What if we see a file with all caps e.g. ID10041? Ideally we'd be case agnostic.

pkmitre: I think we get that with a simple change? Not tested, how about:

        csv_data.rename(columns=lambda c: fieldCaseMapper.get(c.lower(), c), inplace=True)

babraham33: Added this change and got a test case to work!

PR #10, opened 2020-10-20
Support reversion of most recent change to a VA

pkmitre:


babraham33:
Tried to undo two subsequent changes but could only revert the last one. My thinking was that if I revert the latest change, (but there are multiple changes to the record) then I should be able to subsequently undo earlier changes one by one until reaching the original record. Think the problem is that each 'undo' adds a new change to the record history, so subsequent undos will keep 'undoing and redoing' the latest change. Not sure if this is how we'd want to use the undo button, but if so, might want to consider adding logic to handle this functionality.

pkmitre:
With the way version history works we actually expect this behavior: every change to the record is tracked. So, reverting the most recent change needs to add another change to the version history. As currently implemented, this performs the described action of reverting the latest change.

We could change or augment this to support "revert to this version" for every version in the history; that's pretty easy to implement logically, but harder to implement from a UI perspective and I'd be hesitant to add that complexity without an explicit requirement.

babraham33 (approved):

PR #12, opened 2020-11-18
Update/simplify docker stuff and get it working

timster:
This is a big PR but hopefully it's not too scary. Overall the PR attempts to do the following:
- Simplifies docker setup.
- Allows for building "production ready" docker images that can be configured with environment variables.
- Keeps things working locally (running without docker).

I still need to update the Traefik config but this was already getting unwieldy so that can be another (smaller) PR.

I like to try to break down changes by file/directory so here goes...

`config/__init__.py`             
`config/settings/base.py`        
`config/settings/local.py`       
`config/settings/production.py`  
`config/settings/test.py`  
- Simplified base/local/production config a bit.
- Moved some settings from production/local to base that are used for both.
- Removed celery_app import so it doesn't mess up which settings file is used.

`requirements/base.txt`  
`requirements/local.txt`  
`requirements/production.txt`
- Cleaned up base requirements and made sure it had everything needed.
- Removed unused mailgun from production.
- Remove no longer needed local requirements (just use base)

`manage.py`  
`config/wsgi.py`
- Remove `sys.path.append` call since it's not actually needed.

`compose/*` 
- Removed production/local versions so now there is just one version of each image.
- Django now does `collectstatic` on build time instead of runtime.
- Updated import location of `celery_app` due to above config changes.
- Explicitly set `DJANGO_SETTINGS_MODULE` when containers start.

`docker-compose.yml`  
`docker-compose.local.yml`  
- Update docker-compose file to include all images needed.
- Create local override file to expose more ports for local usage.
- Configure images/containers using environment variables.

`local.yml`  
`production.yml`
- Remove docker-compose configs deprecated by those above.

`README.md`
- Updated to reflect all changes above.
- Added instructions for how to run inside docker.


timster:
Updated today with Apache configs and made sure the tests passed (needed to put psycopg in base requirements).

compose/django/Dockerfile
```diff
@@ -14,40 +14,47 @@ RUN apt-get update \
...
 
 # Requirements are installed here to ensure they will be cached.
 COPY ./requirements /requirements
-RUN pip install --no-cache-dir -r /requirements/production.txt \
-    && rm -rf /requirements
+RUN pip install --no-cache-dir -r /requirements/production.txt
 
-COPY ./compose/production/django/entrypoint /entrypoint
+COPY ./compose/django/entrypoint /entrypoint
 RUN sed -i 's/\r$//g' /entrypoint
```
ajbarnes, line 24: Would it be possible to combine some of these RUNs and/or COPYs to [reduce the number of layers](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers)?

README.md
```diff
@@ -125,21 +124,95 @@ Once the prerequisites are available, VA Explorer can be installed and demonstra
...
+## Building/Running in Docker
+
+Django can run locally inside Docker. This will also set up postgres and redis and automatically configure `DATABASE_URL` and `CELERY_BROKER_URL` to use the docker images of postgres and redis.
+
+```
+docker-compose -f docker-compose.yml -f docker-compose.local.yml up django postgres
+```
+
+The server will be running at http://0.0.0.0:5000/
+
```
ajbarnes, line 144: I was able to access the application without issue via these new instructions, however I didn't have a user to login with after. Should we consider adding auto super-user setup or perhaps instructions for setting that up instead?

ajbarnes (commented):
I looked through all the changes and left questions/comments where I thought they made sense, otherwise everything looks good.

I pulled these changes locally and was able to visually confirm the base functionality for the local/production docker setups and successfully ran the tests.

One of my comments addressed logging in to the docker django app after setup. Another experience I encountered while confirming the changes was being able to run the new `docs/test.py` successfully but not being able to view any docs. Do we have sphinx docs to view? If not, should we consider removing them if we don't need them?

compose/django/Dockerfile
timster: Good call. I combined all the sed/chown/chmod commands.

README.md
timster: Yeah, I'll add instructions for creating a super user in docker.

timster:
Resolved merge conflicts.

pkmitre (approved):

PR #15, opened 2020-12-08
Dashboard improvements

babraham33:
This pull request includes the following: 

- improved mapping capabilities (showing adjacent, empty regions in white and adjacent, non-empty regions in gray)
- updated tooltips that use more accurate terminology
- an additional callout box that tracks the number of uncoded VAs
- more robust logic for handling 'no data' situations (both when there is no data in the system, or when a user clicks on a region with no data)

Addresses this [chore](https://www.pivotaltracker.com/n/projects/2455711/stories/175961392)


va_explorer/va_analytics/dash_apps/va_dashboard.py
```diff
@@ -663,65 +690,102 @@ def update_options(search_value, location_json):
...
+        valid_filter = _get_filter_dict(valid_va_df, selected_json, timeframe=timeframe,\
+                                          location_types=location_types, search_terms=search_terms,\
+                                          locations=locations)
+        # filter invalid vas (VAs without COD)
+        invalid_filter = _get_filter_dict(invalid_va_df, selected_json, timeframe=timeframe,\
+                                          location_types=location_types, search_terms=search_terms,\
+                                          locations=locations)    
+        # combine filters into one dictionary to share across callbacks
+        combined_filter_dict = {
+            #"granularity": valid_filter["granularity"], # same across both dictionaries
```
ajbarnes, line 713: Is granularity truly no longer needed?

ajbarnes (approved):
I looked through all the changes and left questions/comments where I thought they made sense, otherwise everything looks good. After addressing those comments, I say LGTM.

I'm happy to report that after pulling the changes I was able to visually confirm the bulleted items and successfully ran the pytests.

As far as general feedback, we might consider truncating the percentages in the "Top 15 Causes of Death" table to 1 or 2 significant figures as right now they appear to be unbound floats. I would also note that at least some of the callout boxes become hidden without being able to scroll to them after screen-width < 1680. We might consider adding an `overflow: scroll` to that container. Additionally, `va_dashboard.py` is a pretty big file and may benefit (in terms of maintainability) from being broken up/out where it makes sense; perhaps treating each table as its own component could be a way to start. These could all  be handled through additional issues outside this PR, though.

va_explorer/va_analytics/dash_apps/va_dashboard.py
```diff
@@ -163,11 +171,13 @@ def load_geojson_data(json_file):
 
 # ============ VA Data =================
 def load_va_data(geographic_levels=None):
-    return_dict = {"data": pd.DataFrame()}
+    np.random.seed(23)
+    return_dict = {"data": {"valid": pd.DataFrame(), "invalid": pd.DataFrame()}}
 
-    valid_vas = VerbalAutopsy.objects.exclude(causes=None).prefetch_related("location").prefetch_related("causes")
+    all_vas = VerbalAutopsy.objects.prefetch_related("location").prefetch_related("causes")
+    #valid_vas = VerbalAutopsy.objects.exclude(causes=None).prefetch_related("location").prefetch_related("causes")
```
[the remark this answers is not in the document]
babraham33: Yeah, no longer necessary. Just removed it.

va_explorer/va_analytics/dash_apps/va_dashboard.py
babraham33: Good catch. I was technically expecting it in update_choropleth (filter_dict.get(granularity, DEFAULT_GRANULARITY), but the get statement defaults to initial granularity so it never threw an error. Anyhow, I uncommented it and it's back in the outermost filter dictionary now.

PR #16, opened 2020-12-10
Pagination

babraham33:
This pull requests adds pagination to the data management tab. By default, only 15 records are loaded at a time, but this can be adjusted as we see fit.

### DIFF
--- a/va_explorer/templates/va_data_management/index.html
+++ b/va_explorer/templates/va_data_management/index.html
@@ -1,5 +1,7 @@
 {% extends "base.html" %}
 
+{% load bootstrap4 %}
+
 {% block title %}Verbal Autopsies{% endblock %}
 
 {% block content %}
@@ -32,6 +34,7 @@
 </table>
 {% if va_data.has_other_pages %}
   <nav aria-label="Page navigation conatiner"></nav>
+    <ul class="justify-content-center">{% bootstrap_pagination va_data size="large"%}</ul>
     <ul class="pagination justify-content-center">
       {% if va_data.has_previous %}
         <li><a href="?page={{ va_data.previous_page_number }}" class="page-link">&laquo; PREV </a></li>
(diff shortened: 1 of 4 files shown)


va_explorer/templates/va_data_management/index.html
```diff
@@ -30,5 +30,18 @@
     {% endfor %}
   </tbody>
 </table>
+{% if va_data.has_other_pages %}
```
timster, line 33: django-bootstrap has a template tag that will render a nice looking paginator in a bootstrap style. Maybe we could try that here? 

https://django-bootstrap4.readthedocs.io/en/latest/templatetags.html#bootstrap-pagination

va_explorer/va_data_management/views.py
```diff
@@ -1,18 +1,36 @@
...
+    va_list = (VerbalAutopsy.objects
+        .prefetch_related("location", "causes", "coding_issues")
+        .order_by("id")
+    )
+
+    # create paginator from query and bucket vas into pages of size page_size
+    paginator = Paginator(va_list, page_size)
+
+    # hone in on active page objects
+    current_page = paginator.page(absolute_page)
```
timster, line 20: We can just call `get_page` instead of `page` then we don't need to do the conversion to `int()` manually above. It's a little safer 

https://docs.djangoproject.com/en/3.1/ref/paginator/#django.core.paginator.Paginator.get_page

babraham33: Thanks for pointing this out. Will change to get_page

va_explorer/templates/va_data_management/index.html
```diff
@@ -30,5 +32,19 @@
...
+  <nav aria-label="Page navigation conatiner"></nav>
+    <ul class="justify-content-center">{% bootstrap_pagination va_data size="large"%}</ul>
+    <ul class="pagination justify-content-center">
```
timster, line 38: Do we still need the custom pagination stuff here? Not sure if it's needed with the bootstrap_pagination or if it's duplicated?

babraham33: Forgot to add a comment about this. I left them in because you have to choose a specific page number with the bootstrap paginator, so I thought it could be nice to have the 'next' and 'previous' functionality once you've honed in on a page range. Can definitely remove if you think it's unnecessary though.

timster (approved):
Looks good to me.

PR #17, opened 2020-12-16
Additional navigation elements for UI

ajbarnes:
This pull request

- Adds a back button to user detail views and a cancel button to user edit forms
- Adds a floating action button partial which toggles additional buttons with navigation features such as back, scroll to top, and scroll to bottom
- Uses the new floating action button on long views/forms within data management
- Adds a font-awesome cdn for icons to compliment the bootstrap cdn

Closes https://www.pivotaltracker.com/story/show/175959547

### DIFF
--- a/va_explorer/static/css/fab.css
+++ b/va_explorer/static/css/fab.css
@@ -1,0 +1,49 @@
+.backdrop {
+	position: fixed;
+	top: 0;
+	left: 0;
+	width: 100%;
+	height: 100%;
+	background: #ECECEC;
+	opacity: 0.6;
+	display: none;
+}
+
+.fab{
+	background: #0070FF;
+	width: 50px;
+	height: 50px;
+	border-radius: 50%;
+	text-align:center;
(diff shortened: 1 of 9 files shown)


babraham33:
I'm having trouble viewing this branch - doesn't show up as an active branch in the main master tree: 
image
It looks like you made this from a forked repo. Any idea how I could view it besides manual download? Hesitant to go that route because I'd have to reset everything.

babraham33:
Looks good, but don't think the buttons are necessary anymore for the data management homepage bc of pagination. Also, I know we're moving away from CDNs, so I'd recommend downloading the fab files and plopping them into static/

timster:
If we go forward with this, would it be possible to go "back" to specific URLs instead of HTTP referrer? The `Referer` header can be spoofed (and possibly exploited), and may not always be present.

ajbarnes:
Closing in favor of !31 which is up-to-date with master and incorporates above feedback.

PR #18, opened 2020-12-23
added bootstrap4 to settings for pagination and removed debugging sni…

babraham33:
…ppet from va_dashboard.py


timster (approved):
looks good

PR #20, opened 2020-12-29
Clean up things related to sites framework.

timster:
As the title says. We don't need to use it. It's pretty straight forward to remove it.


timster:
After some research, we cannot remove the sites app since django-allauth depends on it. So it still exists in INSTALLED_APPS.

However, we can make a few cleanups and remove some unnecessary code.

@ajbarnes I know you asked about this one on the standup yesterday, so tagging you in case you want to take a look.

pkmitre (approved):

PR #28, opened 2021-01-19
Additional Password requirements

ajbarnes:
This pull request

- Adds a PasswordComplexity validator that requires passwords to have at least 1 each of a number, lowercase character, uppercase character, and one of the following special characters `!@#$%^&*()_+-=[]{}|'`
- Adds a configurable PasswordHistory validator that requires users not use any of their previous 12 passwords by default.
- Adds migration for new UserPasswordHistory model
- Adds additional tests for PasswordComplexityValidator and PasswordHistoryValidator

Closes https://www.pivotaltracker.com/story/show/176347269


va_explorer/users/models.py
```diff
@@ -84,3 +89,15 @@ def save(self, *args, **kwargs):
...
         self.username = self.email
         super(User, self).save(*args, **kwargs)
+        if self.original_password != self.password:
+            UserPasswordHistory.remember_password(self)
+
+
+class UserPasswordHistory(models.Model):
+    username = models.ForeignKey(User, on_delete=models.CASCADE)
+    old_password = models.CharField(max_length=128)
+    password_date = models.DateTimeField()
```
timster, line 99: You can add `auto_now_add=True` to this field so you don't have to manually set the value in `remember_password`

va_explorer/users/models.py
```diff
@@ -84,3 +89,15 @@ def save(self, *args, **kwargs):
         # TODO: May need to be changed depending on how username comes in from ODK?
         self.username = self.email
         super(User, self).save(*args, **kwargs)
+        if self.original_password != self.password:
+            UserPasswordHistory.remember_password(self)
+
+
+class UserPasswordHistory(models.Model):
+    username = models.ForeignKey(User, on_delete=models.CASCADE)
```
timster, line 97: Could we call this field `user` to be consistent with other parts of Django?

va_explorer/users/validators.py
```diff
@@ -0,0 +1,111 @@
...
+                )
+
+    def get_help_text(self):
+        return _("Password must not be reused from the previous {} passwords.").format(self.history)
+
+    def _get_last_passwords(self, user):
+        user_password_history = UserPasswordHistory.objects.filter(username_id=user).order_by('id')
+        relevant = user_password_history.count() - self.history
+        # Cleanup older history that we no longer care about if needed
+        relevant = relevant if relevant > 0 else None
```
timster, line 107: This line is not needed. `if relevant` handles the same thing. If relevant is < 0 then `if relevant` evaluates to False anyway.

va_explorer/users/validators.py
```diff
@@ -0,0 +1,111 @@
...
+        if relevant:
+            [entry.delete() for entry in user_password_history[0:relevant]]
```
timster, line 109: Usually not a good practice to build a list for side effects. You can use this to do a bulk delete:

`UserPasswordHistory.filter(pk__in=user_password_history[0:relevant]).delete()`

va_explorer/users/models.py
ajbarnes: Thanks, that's good to know. I've changed this in the latest commit 8e0ab3d

va_explorer/users/models.py
ajbarnes: Yep, happy to make that change to match the convention. Implemented in 8e0ab3d

va_explorer/users/validators.py
ajbarnes: During testing I learned that this line is needed to prevent negative indexing that could occur without the > 0 check and that `UserPasswordHistory.objects.filter(pk__in=user_password_history[0:relevant]).delete()` will throw an `AssertionError` without it. I've left the line in with a comment explaining as much.

va_explorer/users/validators.py
ajbarnes: Good call, I've substituted the logic in the latest commit 8e0ab3d. Thanks.

timster (approved):

PR #29, opened 2021-01-21
Upgrade Django from 3.0 to 3.1

ajbarnes:
This pull request

- Upgrades Django from 3.0.8 to 3.1.5 to take advantage of 3.1 features
- Upgrades `django-allauth`, `django-simple-history`, and `whitenoise` to resolve deprecation warnings caused by the Django version bump
- Upgrades `pytest` and `pytest-django` to ensure testing apparatus validating these changes is up to date
- Resolves TODO in `forms.py` to improve performance of `LocationSelectMultiple` dependent on Django 3.1
- Implements requested changes to `location.js` releating to above TODO.

Closes https://www.pivotaltracker.com/story/show/174481377


timster:
@ajbarnes I tested this morning and made a few commits based on things I found:

- Added new migrations required for Django 3.1
- `descendent_ids` was a function but used like a property. So I changed it to `get_descendant_ids` to be more explicit.
- Switch from `django.contrib.postgres.fields.JSONField` to `django.db.models.JSONField` because `django.contrib.postgres.fields.JSONField` is now deprecated. (see https://docs.djangoproject.com/en/3.1/releases/3.1/#jsonfield-for-all-supported-database-backends)
- Removed pillow dependency (it's heavy and we don't use it anyway) and removed requirements/local.txt since it's not used anymore.

After that I was able to test the UI and I can confirm this does resolve the issue with the user edit page being super slow to load. According to the debug toolbar:

master - 303 queries - about 17 seconds

this branch - 50 queries - about 2 seconds

👍️

timster (approved):

PR #30, opened 2021-01-22
Update views to secure for authenticated user with valid password

timster:
Secure all views to make sure that user is authenticated and has a valid password. If not authenticated, redirect to login page. If user doesn't have a valid password, redirect to set password page.

- Converts all views to class-based views to take advantage of `CustomAuthMixin`.
- Fixed page title in `user_set_password.html`.
- Updates to templates as needed.
- Updates to tests as needed.

Identified a place where 2 GET requests should really be POST requests. Made a few TODO comments so this can be fixed at a later time.

### DIFF
--- a/va_explorer/va_data_management/views.py
+++ b/va_explorer/va_data_management/views.py
@@ -38,6 +38,7 @@ class Show(CustomAuthMixin, DetailView):
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
 
+        context['id'] = self.object.id
         context['form'] = VerbalAutopsyForm(None, instance=self.object)
 
         coding_issues = self.object.coding_issues.all()
--- a/va_explorer/va_data_management/views.py
+++ b/va_explorer/va_data_management/views.py
@@ -1,6 +1,8 @@
+from django.contrib import messages
+from django.contrib.messages.views import SuccessMessageMixin
+from django.core.paginator import Paginator, PageNotAnInteger, EmptyPage
 from django.shortcuts import redirect
 from django.urls import reverse
-from django.core.paginator import Paginator, PageNotAnInteger, EmptyPage
 from django.views.generic import TemplateView, DetailView, UpdateView, ListView
 
 from va_explorer.va_data_management.models import VerbalAutopsy, CauseOfDeath, CauseCodingIssue, Location
@@ -51,11 +53,12 @@ class Show(CustomAuthMixin, DetailView):
         return context
 
 
-class Edit(CustomAuthMixin, UpdateView):
+class Edit(CustomAuthMixin, SuccessMessageMixin, UpdateView):
     template_name = 'va_data_management/edit.html'
     form_class = VerbalAutopsyForm
     model = VerbalAutopsy
     pk_url_kwarg = 'id'
+    success_message = "Verbal Autopsy successfully updated!"
 
     def get_success_url(self):
         return reverse('va_data_management:show', kwargs={'id': self.object.id})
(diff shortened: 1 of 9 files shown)


va_explorer/va_data_management/views.py
```diff
@@ -1,96 +1,92 @@
...
+        context['diffs'] = [new.diff_against(old) for (old, new) in history_pairs]
+
+        return context
+
+
+class Edit(CustomAuthMixin, UpdateView):
+    template_name = 'va_data_management/edit.html'
+    form_class = VerbalAutopsyForm
+    model = VerbalAutopsy
+    pk_url_kwarg = 'id'
```
deastman, line 58: If you import SuccessMessageMixin, you can add something like: 

success_message = "Verbal Autopsy successfully updated!"

See: https://github.com/VA-Explorer/va_explorer/blob/3b5cf5537089ed0ee40271f5dd98164cc0f7866f/va_explorer/users/views.py#L71

va_explorer/va_data_management/views.py
```diff
@@ -1,96 +1,92 @@
...
+        context = super().get_context_data(**kwargs)
+        context['id'] = self.object.id
+        return context
+
+
+class Reset(CustomAuthMixin, DetailView):
+    model = VerbalAutopsy
+    pk_url_kwarg = 'id'
+
+    def render_to_response(self, context):
```
deastman, line 73: Same comment here

import SuccessMessageMixin, add something like:

success_message = "Verbal Autopsy changes successfully reverted!"

va_explorer/va_data_management/views.py
```diff
@@ -1,96 +1,92 @@
...
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context['id'] = self.object.id
+        return context
+
+
+class Reset(CustomAuthMixin, DetailView):
+    model = VerbalAutopsy
+    pk_url_kwarg = 'id'
+
```
deastman, line 72: Same comment here

import SuccessMessageMixin, add something like:

success_message = "Verbal Autopsy changes successfully reverted!"

va_explorer/va_data_management/views.py
```diff
@@ -1,96 +1,92 @@
...
+        latest = self.object.history.latest()
+        if earliest and len(latest.diff_against(earliest).changes) > 0:
+            earliest.instance.save()
+        return redirect('va_data_management:show', id=self.object.id)
+
+
+class RevertLatest(CustomAuthMixin, DetailView):
+    model = VerbalAutopsy
+    pk_url_kwarg = 'id'
+
```
deastman, line 84: Same comments as above.

va_explorer/va_data_management/views.py
timster: Done.

va_explorer/va_data_management/views.py
timster: Done. Couldn't use SuccessMessageMixin here because it's not a form view, but I just imported messages and set the success message directly. Same effect.

va_explorer/va_data_management/views.py
timster: Done. Couldn't use SuccessMessageMixin here because it's not a form view, but I just imported messages and set the success message directly. Same effect.

timster:
@deastman I fixed the tests so I'm done with this one now. Anything else you can think of?

deastman:
@timster This looks good to me!

pkmitre (approved):

PR #32, opened 2021-01-26
Date range filter

babraham33:
This PR adds addresses story [175958444](https://www.pivotaltracker.com/n/projects/2455711/stories/175958444) by adding data management filter functionality enabling the following:
- filter by date range (start date and end date)
- filter by name of deceased
- filter by facility
- see only vas with errors


va_explorer/va_data_management/models.py
```diff
@@ -584,6 +584,15 @@ class VerbalAutopsy(models.Model):
     created = models.DateTimeField(auto_now_add=True)
     updated = models.DateTimeField(auto_now=True)
 
+    # function to tell if VA had any coding errors
+    def any_errors(self):
+        return len([e for e in self.coding_issues.all() if e.severity=='error']) > 0
```
timster, line 589: I think you could simplify this slightly by using [exists](https://docs.djangoproject.com/en/3.1/ref/models/querysets/#django.db.models.query.QuerySet.exists) on the queryset:

`return self.coding_issues.filter(severity='error').exists()`

That way you don't need to loop through all the results.

va_explorer/va_data_management/filters.py
```diff
@@ -0,0 +1,20 @@
+import django_filters
+from django import forms
+
+from .models import *
+from django_filters import DateFilter, CharFilter, DateRangeFilter
+
+from functools import partial
```
timster, line 7: remove unused import

va_explorer/templates/va_data_management/index.html
```diff
@@ -1,15 +1,31 @@
-{% extends "base.html" %}
+<!--{% extends "base.html" %}-->
```
timster, line 1: Should this go back in?

va_explorer/templates/partials/_fab_nav.html
```diff
@@ -0,0 +1,7 @@
+<div class="backdrop"></div>
+<div class="fab child to-top" data-subitem="3"><i class="fa fa-caret-up"></i></div>
+<div class="fab child to-bottom" data-subitem="2"><i class="fa fa-caret-down"></i></div>
+<a href="{{request.META.HTTP_REFERER}}">
```
timster, line 4: We should probably not just blindly redirect back to HTTP referrer.

The Referer header can be spoofed (and possibly exploited), and may not always be present.

Might be better to sanitize it somehow and make sure the referrer (1) exists (2) is a valid location on the current host and not somewhere on a random domain.

Something similar to how Django sanitizes the "next" URL when logging in:

https://github.com/django/django/blob/5fcfe5361e5b8c9738b1ee4c1e9a6f293a7dda40/django/contrib/auth/views.py#L69-L80

va_explorer/templates/base.html
```diff
@@ -19,7 +19,6 @@
     {% block css %}
     <link href="{% static 'vendor/bootstrap-4.3.1/bootstrap.css' %}" rel="stylesheet">
     <link href="{% static 'vendor/select2-4.1.0-beta.1/select2.css' %}" rel="stylesheet">
-
```
timster, line 22: Should this include the fab.js and fab.css?

Do we need to include _fab_nav.html in base.html as well?

va_explorer/va_data_management/models.py
```diff
@@ -621,3 +630,19 @@ class CauseCodingIssue(models.Model):
 
     def __str__(self):
         return self.text
+
+# class VerbalAutopsySummary(models.Model):
```
timster, line 634: Remove commented out model if we're not going to use it.

va_explorer/va_data_management/migrations/0007_delete_verbalautopsysummary.py
```diff
@@ -0,0 +1,16 @@
+# Generated by Django 3.0.8 on 2021-01-26 00:25
```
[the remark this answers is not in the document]
babraham33: yeah sorry about that. Tried a few different approaches and forgot to remove migrations for previous approaches.

va_explorer/va_data_management/models.py
babraham33: good find! Hadn't seen the exists() command before. Will make that update

va_explorer/templates/base.html
babraham33: sorry, the fab files weren't part of this pull request. They accidentally got carried over when I switched branches after reviewing @ajbarnes's navigation PR. Will delete them from my PR

va_explorer/templates/va_data_management/index.html
babraham33: Yes. Added it back in

timster (approved):

PR #33, opened 2021-01-27
Dev user access defaults

deastman:
Addresses the ability to assign a user to an entire country as per: [Pivotal 174481449 ](https://www.pivotaltracker.com/story/show/174481449 )

- Updates the create/update user form to change the UI state upon (1) role selection and (2) geographic access selection. When the Field Worker role is selected, the geographic access radio control is hidden and the form shows only a list of facilities. When a non-field worker role is selected, the form appears as it does in the master branch (the geographic access radio control is available, allowing the user to select either "national" or "location-specific access"; the user can be assigned to provinces, districts, or facilities) Note that Admin, Data Managers, and Data Viewers still default to location-specific access and the user can actively choose national access. 
- Updates the name of the relation on users and locations to user.location_restrictions, indicating that no restrictions in this table means that the user has full, national access (per PK's suggestion)
- Adds support for basic browser-based testing (more tests can be added, but I wanted to include this as a starting point) 

Note: I originally rebased this with master, but the commit history did not look right. Some of the files_changed include files that are already in master, so I reverted. This branch still needs to be rebased with master.


pkmitre (commented):
1. There's a merge commit in the commit history, and what looks like a duplicate commit; I tried to untangle it and pushed my attempt to the `dev_user_access_defaults_rebase` branch, if you'd like to take a look and see whether that new branch correctly captures all the intended changes

2. Do the new tests with selenium require Firefox to be installed? We should discuss whether we want that as a development requirement

pkmitre (changes requested):
Take a look at the `dev_user_access_defaults_rebase` branch to see if that would be appropriate for a new pull request.

deastman:
@pkmitre I will take a look at the rebased branch!

The tests do require Firefox to be installed. Happy to discuss this with the team at a standup!

pkmitre (approved):

PR #36, opened 2021-02-08
Use sqlite in-memory database for tests.

timster:
We might not want to merge this yet.

I just want to submit a PR so the GitHub action runs the tests.


timster:
We should actually be testing on postgresql since that's what we recommend in production. I just wanted to see if tests would work against sqlite in-memory.

PR #41, opened 2021-02-24
Dashboard Responsivity Updates

ajbarnes:
This pull request:

- Moves the search box too its own line helping to frame the map. Adds search icon next to form element to increase prominence 
- Updates the layout to allow dashboard callout boxes content area to shrink/stack when screen width decreases.
- Updates the layout to allow dashboard charts and elements to stack when screen width decreases. Supported down to 480px width
- Links CoD metric up to all dashboard plots
- Bugfix: Incidentally fixes a small issue in the docker-compose config preventing `run_coding_algorithms` from connecting to pycrossVA open port.

Testing:
1. Login and navigate to the dashboard. 
2. Assuming chrome: Either use the developer console's device toolbar to simulate device resolutions or use an extension like [Window Resizer](https://chrome.google.com/webstore/detail/window-resizer/kkelicaakdanhinjdeammmilcgefonfh?hl=en) to change the window size to device presets.
3. Confirm that the dashboard elements respond to screen size changes accordingly down to 480px width.
4. Confirm that CoD dropdown at top of dashboard affects both dashboard plots when changed.

Closes: https://www.pivotaltracker.com/story/show/176870268
Closes: https://www.pivotaltracker.com/story/show/176517861
Closes: https://www.pivotaltracker.com/story/show/176869818


ajbarnes:
@babraham33 All the responsiveness is implemented, but I would also suggest the code be run through a formatter. I've held off on that for now to allow for a better review of the diff (formatting it would add a bunch of diffs not related to my changes) so let me know after you've finished an initial review and I'll submit another commit with formatting changes if that sounds good to you.

babraham33 (commented):
Overall this looks great. A few thoughts: 

1. (minor) Graphs and map seem a bit scrunched in ipad view and some other phone views. Think it's because I set the map column width to 7/12, so when the plots fall below it, they get constrained to that width. Is there a way to change the column width as the width decreases? Or maybe experiment with widening the map to take up 8/12?
2. Might not have been introduced in this PR but the '% of total% button width seems too short to me. 
image
3. I like the new search logo, and definitely more prominent, but not sure about the placement. I think we'd want it higher up so it's one of the first things the user sees, and it's clear that it's a global filter.

smacadam:
Looks good - it would be helpful if the selected region was displayed in the search box when the user clicks on the map, but that can be a future ticket

babraham33 (approved):

PR #42, opened 2021-02-24
Scope verbal autopsies based on user location restrictions

pkmitre:
This pull request

1) Adds a `verbal_autopsies` helper method on the user model that returns just the verbal autopsies the user can access based on their location restrictions (if any)

2) Adds tests for the `verbal_autopsies` helper method

3) Implements location-scoped access control for the verbal autopsy data management functions (viewing, editing, etc)

4) Adds tests for data management access control

5) Incidentally adds the email address of the user making the change to the verbal autopsy change history

Note that this pull request only partially implements location-scoping as needed in the application; for example, it

* Addresses https://www.pivotaltracker.com/story/show/176873370

* Partially addresses https://www.pivotaltracker.com/story/show/176870340, https://www.pivotaltracker.com/story/show/176870399, and https://www.pivotaltracker.com/story/show/176870423

* Does not address https://www.pivotaltracker.com/story/show/176869720

### DIFF
--- a/va_explorer/users/models.py
+++ b/va_explorer/users/models.py
@@ -5,11 +5,12 @@ from django.db.models import ManyToManyField
 from django.urls import reverse
 from django.utils.timezone import localtime
 from django.utils.translation import gettext_lazy as _
+from functools import reduce
 
 # from allauth.account.models import EmailAddress
 # from allauth.account.signals import email_confirmed
 # from django.dispatch import receiver
-from va_explorer.va_data_management.models import Location
+from va_explorer.va_data_management.models import VerbalAutopsy, Location
 
 
 class CustomUserManager(BaseUserManager):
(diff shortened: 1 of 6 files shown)


va_explorer/users/models.py
```diff
@@ -58,6 +59,21 @@ def __init__(self, *args, **kwargs):
 
     location_restrictions = ManyToManyField(Location, related_name="users", db_table="users_user_location_restrictions")
```
deastman, line 61: @pkmitre At standup earlier this week, I asked Matt and Brendan what was meant by Choolwe's description here: "Varied view of system users (modifiable only by Admins)." This bullet is part of one of the tickets I was assigned and comes from Choolwe's Word document. Matt and Brendan mentioned that this refers to geographic scoping. Based on that, it seems like we also need geographic scoping on the Users page.

pkmitre: That makes sense; only admins should see the user management page, but there may be value in allowing "localized" admins. They would see only users for their locations, and when creating a user be limited in what locations they can be assigned to. While I think this is something we want to implement, I suggest that it not be in the highest priority tier of work.

ajbarnes (approved):
I looked over the code and found everything to be looking good and elegant. I might have split the large test methods into multiple tests and had the setUp method handle tree creation, but also consider the one test approach valid since it is testing the single new utility method.

I pulled the changes locally and was able to visually confirm expected functionality of the incidental email and then proper restriction of VAs between admin and fieldworker after assigning the example field worker to a facility. pytest also worked when run locally.

Looks good to merge 👍

va_explorer/users/models.py
ajbarnes: Created [#177124888](https://www.pivotaltracker.com/story/show/177124888) to handle this work in the future.

PR #44, opened 2021-02-28
User role permissions updates

deastman:
**This pull request addresses the following Pivotal stories:**
1. [Confirm or implement administrator privileges](https://www.pivotaltracker.com/story/show/176870340)
2. [Confirm or implement data manager privileges](https://www.pivotaltracker.com/story/show/176870399)
3. [Confirm or implement data viewer privileges](https://www.pivotaltracker.com/story/show/176870423)

**This pull request does not address the following aspects of the above Pivotal stories:**
1. For data manager and viewer: "Varied access to dashboard views (filter modifiable only by Admins)" (Assumption: this means correct geographic access is provided for Dashboard page. This seems to somewhat fall under [#176869720](https://www.pivotaltracker.com/story/show/176869720), though as per Slack discussion, Field Workers should not have access to the Dashboard at the moment)
2. For data manager: "Varied view of system users (modifiable only by Admins)" (Assumption: this means that the correct geographic access is provided for the Users page. It is my understanding that this functionality is on hold for the moment, per Pete's last pull request.)

**This pull request partially addresses the following Pivotal stories in the Icebox:**
1. [Confirm or implement field worker privileges](https://www.pivotaltracker.com/story/show/176870454): A field worker can now view their own profile (under dropdown menu on the right-hand side of the page), but cannot yet view their own verbal autopsies because we have not yet linked identifiers to do so.

**Notes and Questions:** 
1. In adding two tests for the Dashboard, I got an error that out_file was not defined, so this variable was moved within scope to the get method of class DownloadCsv.
2. Should Data Viewers be able to see the Verbal Autopsy change history? I assume yes, but flagging it in case.

**Browser Testing:**
1. Run migrations
2. Rerun initialize_groups.py: `python manage.py initialize_groups` (Likewise, Choolwe's team must also rerun initialize_groups.py for the change in permissions to take effect.) 
2. Seed demo users: `python manage.py seed_demo_users`
3. Log in as: data_manager@example.com, data_viewer@example.com, field_worker@example.com, and an admin user (run `python manage.py seed_admin_user` as per README if necessary)
4. Visually confirm that all user groups have correct permissions


va_explorer/templates/va_data_management/index.html
```diff
@@ -46,7 +46,9 @@
         <td>{{ va.cause }}</td>
         <td>{{ va.warnings }}</td>
         <td>{{ va.errors }}</td>
-        <td><a class="btn btn-primary" href="{% url 'data_management:show' id=va.id %}">View</a></td>
+        {% if perms.va_data_management.view_verbalautopsy %}
```
timster, line 49: Do we actually need to check this here? The view itself is protected by a PermissionRequiredMixin so I don't think you could even get to index.html without alread having this permission.

timster (commented):
Looks good to me. I just had a minor comment about checking permissions.

va_explorer/templates/va_data_management/index.html
deastman: Great point! We don't need it here. I will remove it. :-)

timster (approved):

PR #46, opened 2021-03-02
Update user session permissions

smacadam:
Addresses this [ticket](https://www.pivotaltracker.com/story/show/176870587) Previously, user permissions were not updated until the user logged out and signed back in. Discussed with Pete and rather than force a logout, I updated the user's session on a successful role update. 

Testing:
- Sign in as an admin and 2nd user (any role) in two separate windows
- Use the admin account to modify the 2nd user's role
- Refresh the 2nd user's page and make sure the new permissions are applied
- Navigate to a new page and make sure the 2nd user's new permissions are applied
- Check the admin still has all expected permissions
- Edge case: make the 2nd user an admin and use the 2nd user to edit their own  role
  - should get a success message and potentially a 403 depending on the new role
  - navigate to the data management page and make sure new permissions are applied


ajbarnes (approved):
Everything looks good on the small code review. Local pull confirms visual and automated tests. LGTM!

PR #47, opened 2021-03-03
Add metrics and actions to home page

pkmitre:
This pull request:

* Adds some basic metrics and small charts to the home page for the currently logged in user, showing counts of recently collected, coded, and uncoded records for the regions they can access
* Adds a worklist of records that have not been coded that allows the user to immediately address any errors, and a link to view all records with errors
* Refactors record listing into a common shared template
* Adds basic tests

Notes:

* This implementation currently uses the "created" field for categorizing records by date; this should likely use the date the VA is completed, which we don't currently import (note that we do import the date of death, which we display in record listings, but that's not appropriate to use for data collection stats); suggest we implement this in a future PR (this has been captured as https://www.pivotaltracker.com/story/show/177195813)

* There has been some discussion about the best approach for adding graphs to this page (e.g. combining the three current graphs into one graph with different colors); suggest that we defer making further changes until we can get user input on what questions they'd like answered, and allow that to drive graphing and display decisions (this has been captured as https://www.pivotaltracker.com/story/show/177195743)

* The idea of running (or re-running) the coding algorithms via a button on the home page has been suggested; suggest implementing that in a future PR (this has been captured as https://www.pivotaltracker.com/story/show/177195813)

HomePageMetrics

### DIFF
--- a/va_explorer/home/tests/test_views.py
+++ b/va_explorer/home/tests/test_views.py
@@ -1,0 +1,38 @@
+import pytest
+from django.test import Client
+from va_explorer.users.models import User
+from va_explorer.va_data_management.models import VerbalAutopsy
+from va_explorer.tests.factories import VerbalAutopsyFactory, LocationFactory
+
+pytestmark = pytest.mark.django_db
+
+# Get the index and make sure 
+def test_index(user: User):
+    client = Client()
+    client.force_login(user=user)
+    va = VerbalAutopsyFactory.create()
+    response = client.get("/")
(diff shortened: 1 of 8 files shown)


ajbarnes (approved):
Code review seemed like an elegant solution with appropriate tests and setup scripts for potential demo. Pulling the changes locally also visually confirmed expected functionality of the new metrics page, changes to the about page, and working [py]tests - both old and new. I say LGTM!

PR #51, opened 2021-03-09
Add configurable tooltips to callout boxes

ajbarnes:
This PR addresses this [Pivotal #176987730](https://www.pivotaltracker.com/story/show/176987730) - "Either re-phrase certain "misleading" labels or provide info tooltips with context" sourced from user feedback.

Verification:
- login and navigate to the dashboard
- hover over the info icon for each callout box
- tooltip with relevant info should display on hover and disappear after hover


babraham33 (commented):
Looks great! Just have two thoughts: 
1. consider lightening the background color of info boxes to light gray. Black draws the user to them first instead of their values. Something like this looks good: 
image

2. For regional rep, change '% of regions within geography' to something like "% of regions with data within geo" or "% of actively reporting regions within geo". Something that implies they are active and are submitting VAs.

babraham33 (approved):

PR #52, opened 2021-03-10
Dashboard optimizations

babraham33:
This PR includes various optimizations to improve dashboard rendering time and enable deployment with multiple gunicorn workers. Specifically: 
- Time period filter logic changed so it doesn't trigger new DB query upon change 
- Unnecessary callout removed that prevented the map from updating properly with multiple workers
Addresses [this issue](https://www.pivotaltracker.com/n/projects/2455711/stories/176872188)

**To Test**: 
1. run app via docker compose (`docker-compose up --build`)
2.  browse to localhost:5000 and go to dashboard
3. confirm that map renders when new district is searched for or clicked on map, and that all callout boxes and plots update accordingly
4. confirm that changing the time filter works and that page now renders in < 2 seconds

### DIFF
--- a/va_explorer/va_analytics/dash_apps/va_dashboard.py
+++ b/va_explorer/va_analytics/dash_apps/va_dashboard.py
@@ -619,14 +619,9 @@ designated as "expanded"
     ],
     [Input(component_id="hidden_trigger", component_property="children")],
 )
-<<<<<<< HEAD
 
-def init_va_data(timeframe="All", **kwargs):
+def init_va_data(hidden_trigger=None, **kwargs):
     res = load_va_data(kwargs['user'])
-=======
-def init_va_data(hidden_trigger=None):
-    res = load_va_data()
->>>>>>> 9772e88bee03e77ccd9af2610dcf2505723e5e94
     valid_va_data = res["data"]["valid"].to_json()
     invalid_va_data = res["data"]["invalid"].to_json()
     locations = json.dumps(res.get("locations", []))
@@ -765,11 +760,8 @@ def _get_filter_dict(
     plot_ids, plot_regions = list(), list()
     
     filter_dict = {
-<<<<<<< HEAD
+
         "geo_filter": any(map(lambda x: len(x) > 0, [restrictions, selected_json, search_terms])),
-=======
-        "geo_filter": (selected_json is not None) or (len(search_terms) > 0),
->>>>>>> 9772e88bee03e77ccd9af2610dcf2505723e5e94
         "search_filter": (len(search_terms) > 0),
         "plot_regions": [],
         "chosen_region": "all",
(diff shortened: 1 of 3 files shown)


va_explorer/va_analytics/dash_apps/va_dashboard.py
```diff
@@ -672,6 +688,15 @@ def filter_data(
             json.loads(location_types) if location_types is not None else location_types
         )
         disable_timeframe = False
+        
+        # check if user has geoscoping restrictions. If so, filter down to allowed locations
```
timster, line 692: I think you could take advantage of values_list here, then you don't need the try/except:

```kwargs["user"].location_restrictions.values_list("name", flat=True)```

File: va_explorer/va_analytics/dash_apps/va_dashboard.py [hunk 2]
```diff
@@ -712,66 +739,95 @@ def filter_data(
...
     if filter_dict["geo_filter"]:
-        # first, check if user searched anything. If yes, use that as filter.
+        # first, check if geo filter is for location restrictions
+        if len(restrictions) > 0:
+            # TODO: make this work for more than one assigned region
+            granularity = locations.get(restrictions[0], granularity)
+            # no need to filter data, as that's already done in load_data
+            chosen_region = restrictions[0]
+            
+            location_obj = Location.objects.get(name=chosen_region)
```
timster, line 780: Is it possible there could be more than one location with the same name? Would it be possible to key this off Location ID instead?

va_explorer/va_analytics/dash_apps/va_dashboard.py
```diff
@@ -1057,7 +1103,13 @@ def update_choropleth(
                     dragmode="pan",
                 )
                 # additional styling
-                config = graph_config
+                config = LOOKUP['graph_config']
+                
+                # if geo restrictions in place, disable clicking
+                if len(kwargs["user"].location_restrictions.all()) > 0:
```
timster, line 1109: ```kwargs["user"].location_restrictions.exists()``` to be a little more efficient.

va_explorer/va_analytics/dash_apps/va_dashboard.py
babraham33: Good point. It's definitely possible for location restrictions - changed the logic to pull in both name and ids. However, it's trickier for location-search and map-click based filtering, because those locations are all name based. Will need to give those some thought but problems for another day.

timster (approved):

va_explorer/va_analytics/dash_apps/va_dashboard.py
```diff
@@ -641,6 +655,7 @@ def update_options(search_value, location_json):
     [
         Output(component_id="filter_dict", component_property="children"),
         Output(component_id="timeframe", component_property="disabled"),
+#        Output(component_id="bounds", component_property="children")
```
ajbarnes, line 658: Do we see this commented out line as being useful going forward? If not, I'd recommend removal.

va_explorer/va_analytics/dash_apps/va_dashboard.py
```diff
@@ -904,13 +948,14 @@ def update_choropleth(
...
     # first, see which input triggered update. If granularity change, only run
     # if value is non-empty
-    context = dash.callback_context
-    trigger = context.triggered[0]
-    if trigger["prop_id"].split(".")[0] == "view_level" and trigger["value"] == "":
-        raise dash.exceptions.PreventUpdate
+#    context = dash.callback_context
+#    trigger = context.triggered[0]
+#    if trigger["prop_id"].split(".")[0] == "view_level" and trigger["value"] == "":
+#        raise dash.exceptions.PreventUpdate
```
ajbarnes, line 952: Similarly, should this and any other commented out code be kept?

PR #56, opened 2021-03-17
Empty map district or province

smacadam:
This PR addresses this [ticket](https://www.pivotaltracker.com/story/show/177305979) where selecting empty districts/provinces on the map caused errors. This was resolved by stopping updates to the map if the user clicks an empty data region. This also caused errors in the plots because they did not handle the case where the data is empty. The map also needed to handle empty data after other filters like time period are applied. 

Testing:
- Go to the dashboard view and click an empty data region on the map, the map should not update and the plots on the right and call out boxes should be updated to display no data
- Reset the dashboard (note there is a bug with reset button that's addressed in another PR so you'll have to refresh the page) and set the time period filter to "today", the map should update and have all white "No Data" regions, the plots and callout boxes should also update to show no data


ajbarnes (approved):
Code review looked good and pulling the changes locally confirmed the same. Instructions worked as expected, including pytest, with `dashboard-optimizations` actually fixing the reset bug. LGTM

PR #63, opened 2021-03-23
Add ability to import from ODK

timster:
Adds the ability to import records from an ODK server. Adds directions to the README for how to use the new functionality.

Tested against a locally running ODK server and also contains unit tests with mocked data. Added requests-mock to be able to mock responses from an ODK server in unit tests.

The error handling is not super robust. If there is an error communicating with the ODK server, it will bubble up to the user. That is probably okay for now since this is a back-end command and not exposed through the front-end (yet).

## Files Changed

`modified: README.md`
- Added instructions for running new functionality.

`modified: requirements/base.txt`
- Added requests-mock dependency.

`new file: va_explorer/va_data_management/utils/odk.py`
- Contains core functionality for communicating with ODK server, authenticating, and downloading records.

`new file: va_explorer/va_data_management/management/commands/import_from_odk.py`
- New management command for importing from ODK.

`new file: va_explorer/va_data_management/tests/test_odk.py`  
`new file: va_explorer/va_data_management/tests/odk-data.csv`  
`new file: va_explorer/va_data_management/tests/odk-data.json`
- Added tests to cover all ODK import functionality.
- The csv and json files are mock responses for downloading form that were too big to inline in a python file (see the test for how they are used).

`modified: va_explorer/va_data_management/utils/loading.py`
- Slight modification to account for instanceID column case sensitivity.


pkmitre (changes requested):
With the current implementation it seems that whenever records are imported from ODK every record is pulled and written to the database, even if already pulled previously. I think that before rolling this out we should update to only import records that haven't already been imported.

 whenever data is imported

timster:
> With the current implementation it seems that whenever records are imported from ODK every record is pulled and written to the database, even if already pulled previously. I think that before rolling this out we should update to only import records that haven't already been imported.

@pkmitre I have updated the logic in loading.py to account check for existing VA with matching instanceid for each incoming row. Updated import_from_odk.py and load_va_csv.py to handle this (they will now report number of 'ignored' records). Updated corresponding tests in test_loading.py and test_odk.py.

PR #64, opened 2021-03-24
Add va_username for fieldworker

deastman:
This pull request partially implements [Pivotal Story, Ability to tie VAs to interviewers](https://www.pivotaltracker.com/story/show/177311007). I am breaking this ticket up into multiple pull requests.
This pull request implements the first part of this story:  "Each VA Explorer interviewer user account should be able to be tied to a username (we may eventually want to support more than one username, but start with one) on the user management page"

**This major components of this pull request are:**
* Adds an optional "username" field (va_username in the database) to the user creation and update forms that is only shown when the Field Worker role is selected
* Adds a "username" field to the UserDetail page, that is shown only for Field Worker users
* Adds the model VaUsername that has a ForeignKey to the user and a va_username field (this field will be joined on the username field in va_data_management_verbalautopsy ) Note: I am not calling this field "username" in the model to avoid confusion with the username field that exists on the user model and which comes automatically from django.contrib.auth.
* Implements the va_username field as one-to-one with the User through logic in users/forms.py. This logic can be updated to accommodate having more than one va_username per Field Worker if needed.
* Updates tests to test for correctness

**To test this functionality in the UI:**
1. As an admin user, confirm that the Username field is only visible on the User Create and Update forms when the Field Worker role is selected
2. As an admin user, confirm that the Username field is optional when creating a Field Worker
3. As an admin user, confirm that the Username field is optional when updating a Field Worker
4. As an admin user, confirm that you can assign a Username to a Field Worker and that the Username appears on their UserDetail page
5. As an admin user, confirm that you cannot assign the same Username to two different Field Workers
6. As an admin user, confirm that updating a Field Worker with a preexisting Username and leaving the Username field blank on update, removes the Username


va_explorer/templates/users/user_detail.html
```diff
@@ -42,7 +42,25 @@
         {%  endfor %}
     </div>
 </div>
-<br/>
+{% comment "Note that Django does not support filtering of query sets in templates" %}
+{% endcomment %}
+{% for group in user.groups.all %}
+  {% if group.name == "Field Workers" %}
```
timster, line 48: Wondering if we should have a generic `is_field_worker()` method (or `@property`) on the user so we don't need to loop and check for the string everywhere?

va_explorer/users/forms.py
```diff
@@ -247,6 +294,28 @@ def save(self, commit=True):
             user.location_restrictions.set(location_restrictions)
             user.groups.set([group])
 
+            # TODO: Update this if we are supporting more than one username; for now, allow only one
```
timster, line 297: Do you think it would be worthwhile to have a `set_va_username(*usernames)` method on User model that encapsulates all the logic for adding/removing usernames? This would allow us to use/set the username outside of a form easily if we need to do so.

va_explorer/users/views.py
```diff
@@ -104,6 +114,11 @@ def get_initial(self):
                 self.get_object().location_restrictions.filter(location_type="facility")
         )
 
+        # TODO: Update this if we are supporting more than one username;
```
timster, line 117: See other comment about encapsulating this logic on the user model.

va_explorer/users/views.py
```diff
@@ -56,6 +57,12 @@ class UserDetailView(CustomAuthMixin, UserDetailViewMixin, DetailView):
     login_url = reverse_lazy("account_login")
     model = User
 
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context['va_usernames'] = VaUsername.objects.filter(user_id=self.object.id)
```
timster, line 62: Since it's a relation to user, you could do `user.vausername_set.all` in the template without needing to set this context variable.

va_explorer/users/views.py
```diff
@@ -104,6 +114,11 @@ def get_initial(self):
                 self.get_object().location_restrictions.filter(location_type="facility")
         )
 
+        # TODO: Update this if we are supporting more than one username;
+        #  For now, we only ever allow one, so we will display one
+        va_username_for_user = VaUsername.objects.filter(user_id=self.get_object()).first()
```
timster, line 119: You can do `self.get_object().vausername_set.first()`

va_explorer/users/forms.py
deastman: @timster Yes! I think that would be great! I will implement your suggestion.

timster (approved):
Looks good. I tested the functionality and it works as expected.

PR #69, opened 2021-03-30
Cod group plotting 

babraham33:
Added the following functionality to the dashboard:

- plot multiple COD groups in COD analysis tab
- plot multiple CODs and/or groups in VA trends tab
- if a COD is chosen from the global dropdown (top of the dashboard), that COD is highlighted/bolded in group plots
- minor language improvements

Also added COD groupings file that can be extended if new groupings are needed (see`va_analytics/dash_apps/dashboard_data/cod_groupings.csv`)

This PR addresses:
- https://www.pivotaltracker.com/n/projects/2455711/stories/176869793
- https://www.pivotaltracker.com/n/projects/2455711/stories/177209479
- https://www.pivotaltracker.com/n/projects/2455711/stories/177342448


babraham33:
> For this ticket, https://www.pivotaltracker.com/n/projects/2455711/stories/177209479 I tried comparing top 10 COD overall vs. top 10 COD by gender to see if the result made sense. But the top 10 by gender only showed 9 COD instead of 10. Is this expected?
> image

Great catch! I was accidentally filtering out the last COD because I had an extra row total in the pivot table. Got rid of that, so the issue should be fixed now

babraham33:
> When plotting the COD in the VA trends tab, the COD drop down appears to be empty. If I start typing, the options appear, but it was a bit confusing at first. Would it make sense to set that value to "All Causes" by default similar to the COD tab?

Yes, definitely. Made that the default behavior now.

smacadam (approved):

PR #70, opened 2021-03-31
Field worker va views

deastman:
This pull request addresses bullet 2 of [Ability to tie VAs to interviewers](https://www.pivotaltracker.com/story/show/177311007) (the first bullet is already in master)

Specifically, a Field Worker is only permitted to see Verbal Autopsies that they have created, as determined by the username field on the Verbal Autopsy and its join on the va_usernames table. 

The pull request also updates the home page template, va_data_management index template, and va_data_management show template to account for elements the group thought it made sense to hide for the time being from the Field Worker users (discussed during standup on 3/26). 

Hiding these data and form controls are a temporary solution until the team decides if the Field Worker should have access to more robust actions on the Verbal Autopsy (i.e., making changes to their own verbal autopsies, which may be held for approval by an administrator). 

**Bullet 3 of this story is not yet implemented:** 
"We should consider a) listing the interviewer when we list VAs and b) allowing VAs to be filtered by interviewer" (potentially make this a new ticket?)

**This story is related to:**
[176870454](https://www.pivotaltracker.com/story/show/176870454) (duplicative)
[177364238](https://www.pivotaltracker.com/story/show/177364238) (enhancement/next steps)

**To test this functionality via the user interface:**
1. Edit one Verbal Autopsy in the system to add a username (e.g., "user123")
2. Create a field worker with the same username, "user123"
3. Log in as this field worker and confirm that this field worker user only has access to the Verbal Autopsy with username "user123"


va_explorer/va_data_management/views.py
```diff
@@ -23,11 +23,17 @@ def get_queryset(self):
         # Restrict to VAs this user can access and prefetch related for performance
         queryset = self.request.user.verbal_autopsies().prefetch_related("location", "causes", "coding_issues").order_by("id")
 
+        # TODO: For now, we are not displaying the filters for the Field Worker on the VA index page,
+        # since many do not apply to them. This prevents data passed through the params from being
+        # passed to the VAFilter
+        if self.request.user.is_fieldworker():
+            self.filterset = VAFilter(data=None, queryset=queryset)
```
timster, line 30: You do an early return here and just do `return queryset` instead since it doesn't need all the stuff below it for filterset.

But it doesn't matter a whole lot and is kinda nit-picky.

va_explorer/users/models.py
```diff
@@ -63,16 +63,18 @@ def __init__(self, *args, **kwargs):
...
-        location_sets = [Location.get_tree(location) for location in self.location_restrictions.all()]
-        locations = reduce((lambda set1, set2: set1 | set2), location_sets)
-        # Return the list of all verbal autopsies associated with that query set of locations
-        return VerbalAutopsy.objects.filter(location__in=locations)
-      else:
-        # No location restrictions, which implies access to all data
-        return VerbalAutopsy.objects.all()
+        if self.is_fieldworker():
+            return VerbalAutopsy.objects.filter(username__in=self.vausername_set.all().values_list('va_username'))
+        elif self.location_restrictions.count() > 0:
```
timster, line 68: No need for `elif`  here (could use regular `if`) since the above line returns. But again this is nit-picky.

va_explorer/va_data_management/views.py
deastman: @timster Thank you! I tried a couple of combinations of things here, but couldn't get it to work correctly without setting the filterset in this way. Do you think this is okay as is? I might be missing something totally obviously here.

PR #72, opened 2021-03-31
Remove random assignments

smacadam:
This PR removes randomization for the following fields: Id10023(date of death), ageInYears and age_group, location
- Id10023: (date of death): if not parseable as a date, an error is recorded
- ageInYears is parsed from the ageInYearsField
- age_group checks several fields for information to determine age_group, if no relevant data is found age_group is set to "unknown"
- location checks Id10058 for a known facility. If not a known facility, it checks for a username assigned to the VA and grabs the user's first assigned location. If no username, uses Null location (called Unknown in db)(see [ticket](https://www.pivotaltracker.com/story/show/177588225) 
- The "validation" is broken into 3 parts
- VA model clean: the clean function derives the location field
- VA edit form clean: formats Id10023 as a human-readable date
- validate.py takes in a list of va records and updates all errors and warnings for these fields so they are visible in the data management
- The only unknown that caused the dashboard to break was Id10023. As such, if Id10023 is unknown, the record is filtered out of the dashboard records, but all other unknowns are included in the dashboard. Here's a [ticket](https://www.pivotaltracker.com/story/show/177588225) to check whether any other unknowns should be filtered out of the dashboard

For development purposes, this PR also adds ability to randomly assign locations to VAs if desired by setting the `random_locations` flag to `True` when calling `load_va_csv`. This is done by randomly assigning a field worker in the system to each VA. If there aren't any field workers with location restrictions, synthetic field workers are generated for each facility before the assignment process. 

Testing
- make sure you have mock data with Id10023 date values
- reset your db 
source venv/bin/activate
python manage.py reset_db
python manage.py migrate
python manage.py initialize_groups
python manage.py seed_admin_user admin@example.com --password Password1
python manage.py seed_demo_users
python manage.py load_locations zambia_locations.csv
python manage.py load_va_csv mock_data_with_Id10023_values.csv 
         **Note**: If VAs aren't assigned to field workers yet, add `--random_locations True` to command line arguments
python manage.py run_coding_algorithms

- check out the data management page, make sure the "dates" for each va match the values in the csv Id10023 field
- Test Id10023:
  - edit the first record with a known cause of death, change Id10023 to "dk" and save
  - the record should save successfully, it will appear on the last page since the va's are sorted by date
  - navigate to the dashboard to make sure there are no errors displaying the va's
  - return to the data management page

- Test ageInYears
  - edit a record with a known cause of death
  - check the ageInYears value, if it is set, clear the data and save
  - check the va's errors, there should be a warning that says ageInYears is not set
  - set ageInYears, save, and verify the error is gone
- Test age_group
  - edit a record with a known cause of death
  - clear data from fields: age_group, isNeonatal1, isChild1, isAdult1m and ageInYears and save
  - check the va's errors, there should be a warning that says age_group cannot be determined
  - set any of the fields above and save, the ageInYears and age_group warnings should be gone
- Test location
  - Copy a facility name from the data management view
  - Edit a VA with a different facility, set Id10058 to the facility name and save
  - The facility in the list view should be updated
  - Create a field worker and assign them a facility
  - Once the field worker is created, edit a VA
  - Set the Id10058 field to "home" or "ref", and set the username field to the username of the user you just created 
  - save the record, the facility should be updated to the facility assigned to the field worker


pkmitre (changes requested):
This pull request looks good and makes sense to me. I've rebased on master and integrated these changes with the changes made for ODK import; someone should review the changes I made before merging. I also suggest holding off on merging this pull request until the remaining issues in https://www.pivotaltracker.com/story/show/177588225 (I think supporting a null location could make sense) and https://www.pivotaltracker.com/story/show/177588249 are addressed.

babraham33:
@pkmitre I've reviewed your changes and they look good. I also added logic to support a null location (addressing (https://www.pivotaltracker.com/story/show/177588225 )) and added a new function to assign locations via the username of randomly chosen, synthetic field workers. By default, this behavior is disabled, but it can be enabled by passing `random_locations=True` to `load_records_from_dataframe`. Lastly, I improved the date handling to support a variety of formats instead of just a single one. Between Sarah's changes and my most recent changes, I think we've addressed https://www.pivotaltracker.com/story/show/177588249 as well. Records without a location or username are set to the null location, and we have logic in the dashboard to handle unrecognized age groups. Let me know what you think when you get the chance.

pkmitre (approved):

PR #74, opened 2021-04-06
removed field worker calculation from dashboard until we can calculat…

babraham33:
…e properly


ajbarnes (approved):
This seems like a reasonable step for the pilot to avoid confusing the user.

PR #75, opened 2021-04-09
Update README to include attribution line under license

mboyas-mitre:
Added attribution line under license

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,6 +294,8 @@ Releases are documented in the [CHANGELOG]().
 
 Copyright 2020-2021 The MITRE Corporation
 
+The source of this information is the Data for Health Initiative, a joint project of the CDC Foundation and Bloomberg Philanthropies.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -294,6 +294,8 @@ Releases are documented in the [CHANGELOG]().
 
 Copyright 2020-2021 The MITRE Corporation
 
+The source of this information is the Data for Health Initiative, a joint project of the CDC Foundation and Bloomberg Philanthropies.
```
pkmitre, line 297: Is this language they provided? "The source of this information" reads a bit odd given that it's describing source code (yes, it is "information", but not usually referred to that way...).

mboyas-mitre: Yes, it's provided though I do think we can tweak a little if we want to do so.   Might be easiest to just leave as is though for now and we can discuss specifics later?

timster (approved):

PR #76, opened 2021-04-15
Fix DJANGO_DEFAULT_FROM_EMAIL

timster:
enables setting DJANGO_DEFAULT_FROM_EMAIL from environment variable from docker

va_explorer.org is not valid (domains cannot contain underscores) so the mail server would reject it.


mboyas-mitre:
Seems fine to me!

mboyas-mitre (approved):

PR #78, opened 2021-04-19
added logic to parse locations from hospital field and new odk date format

babraham33:
This PR adds logic to assign location based on hospital field and improves date date handling to support new odk date format.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -10,6 +10,10 @@ pytz==2020.1
 redis==3.5.3
 requests==2.25.1
 
+# String matching
+fuzzywuzzy==0.18.0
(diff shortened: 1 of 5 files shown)


mboyas-mitre (commented):
I just took a look, and I’d also like it if @ajbarnes could review as well.

My one thought here, which I will also add to the PR, is that I feel like the regexs could be somewhat variable given changing data - it may be good, if even for pilot purposes, to append to a log where we have import date, original field, and extracted field?  so we can do some checks over the course of the pilot to make sure this is doing everything we want and nothing else?

mboyas-mitre (approved):

PR #81, opened 2021-04-29
Add ability to redact fields that contain PII

timster:
This PR adds the ability to redact certain fields for users who do not have permissions to view PII.

Added a new `view_pii` permission to control this. This permission is added by default to Admins and Data Managers. It can be added explicitly to any other user who needs it.

In templates, added a `pii_filter` template tag to redact fields as needed (see usage in templates).

In views or other areas in python, check fields against the list of PII_FIELDS and handle as needed.

For editing a record, if user has `edit_verbalautopsy` but not `view_pii` then none of the fields that are redacted will show in the form. See how it works in va_data_management/forms.py. This will not be the case with any of the default groups since the only groups that have `edit_verbalautopsy` also have `view_pii`. Just something to be aware of.

On the user create/udpate screens, added a "Can View PII" checkbox for the user.  If the user's group includes the permission, you will not be able to uncheck it for the user. If the user's group does not include it, you can add it specifically for the user.

Also fixed the permission for download_data on the download CSV view and fixed/cleaned up some tests.

## Permission

`va_analytics/models.py`  
`va_analytics/migrations/0002_update_dashboard_permissions.py`
- Add new view_pii permission.
- Fix description for download_data permisssion.

`va_analytics/views.py`  
`va_data_management/tests/test_views.py`
- Fix permissions so downloading data requires download_data permission.
- Redact fields in downloaded file as needed.

`users/management/commands/initialize_groups.py`
- Add permission for view_pii.
- Fixed the way permissions are looked up (previously it was broken for a few like download_data).

`users/models.py`
- Added `can_view_pii` helper property/setter for a user.

## Viewing/Editing Verbal Autopsy

`va_data_management/models.py`
- Add list of fields that need to be redacted.

`home/views.py`
- Fixed issue with showing date (to match Unknown like we do in other places).

`templatetags/va_explorer_tags.py`
- Added new `pii_filter` tag to redact fields if needed.

`templates/partials/_va_table.html`  
`templates/va_data_management/show.html`
- Use `pii_filter` to redact fields.

`va_data_management/forms.py`
- Remove PII fields if user does not have permission to view PII.

`va_data_management/views.py`  
- Pass the ability to edit PII to the verbal autopsy form.
- Only show filters if user can view PII (filters only apply to redacted fields).

`templates/va_data_management/index.html`
- Only show filters if user can view PII (filters only apply to redacted fields).

## Viewing/Editing Users

`users/forms.py`
- Added `UserCommonFields` to store fields that are common across user create/update forms.
- Added `GroupSelect` widget for group selection that has data-view-pii attribute (used by location.js).
- Added `view_pii` checkbox on user create/update forms.
- Save `user.can_view_pii` when users are created/updated.

`users/views.py`
- Set initial `view_pii` for user update form.

`static/js/location.js`
- Added JavaScript for enabling/disabling the View PII checkbox based on group selection.

`templates/users/user_detail.html`
- Add display for user "Can View PII".

## Tests

`va_analytics/tests/test_views.py`
- Fix a couple broken tests.
- Update/add tests to work with PII functionality.


ajbarnes (approved):
Pulled locally and was able to build/run pytest no issues. Testing around in the UI with the various different users was mostly good as well. I was able to confirm all the changes like redactions and data download redactions.

It may be easy to overlook, so I'll note that it was important to re-run `./manage.py initialize_groups` as well as `./manage.py migrate` since both were affected by this PR.

I ran into a bug while attempting to give the `view_pii` permission to the demo data_viewer user but wasn't able to determine if that was related to this PR's scope so I created an issue with the stack trace to track: [#177988365](https://www.pivotaltracker.com/story/show/177988365)

PR #82, opened 2021-04-30
fixed location validation issues preventing user from saving changes …

babraham33:
fixed location validation issues that were adding faulty errors and preventing user from editing va records. 

To test: 
  import some va data, run the server and try editing a record (ex. minor tweak like first name change) and saving changes. You should get confirmation that the change went through


va_explorer/va_data_management/utils/validate.py
```diff
@@ -74,36 +75,31 @@ def validate_vas_for_dashboard(verbal_autopsies):
...
-                locations = va_user.user.location_restrictions
-                if  locations.count() == 0:
-                    issue_text = "Error: fields username, cannot determine map location, username has no assigned facilities."
-                    severity = "error"
+        known_location = Location.objects.filter(name=va.location).first()
+        if not known_location:
+            # try re-assigning location using location logic described in loading.py
+            va = assign_va_location(va)
+            if va.location:
+                if va.location == "Unknown":
```
timster, line 83: I think `va.location` is an instance of `Location` at this point? So should it check `va.location.name` instead?

Or maybe add as a helper on VerbalAutopsy like `va.has_null_location` so we don't need to check for a string here.

va_explorer/va_data_management/utils/validate.py
```diff
@@ -74,36 +75,31 @@ def validate_vas_for_dashboard(verbal_autopsies):
...
+                    issue_text = "Warning: location field (parsed from hospital):, the location provided was not a known facility. Defaulted to Null Location (Uknown)"
+                    severity = "warning"
                     issue = CauseCodingIssue(verbalautopsy_id=va.id, text=issue_text, severity=severity, algorithm='', settings='')
-                    issues.append(issue)  
+                    issues.append(issue) 
+                else:
+                    known_location = Location.objects.filter(name=va.location).first()
```
timster, line 89: Same comment here - isn't `va.location` an instance?

timster (approved):

va_explorer/va_data_management/utils/validate.py
babraham33: Good catch! I think you're right. I'll use .name for comparison in both cases

timster:
Why do we go through the trouble of setting up / searching for a location with name "Unknown"?

Could we just make location a nullable field?

If it's a nullable field you can do things like this:

`va.location = Location.objects.filter(name='whatever').first()`

And if there is not a location with name `whatever` then it will get assigned null.

Then you can assume any VA with a null location is "Unknown".

Seems like that could simplify a lot of logic - especially what's in `set_null_location` as you could just set `va.location = None`

But maybe I am missing something that prevents us from doing that.

babraham33:
> Why do we go through the trouble of setting up / searching for a location with name "Unknown"?
> 
> Could we just make location a nullable field?
> 
> If it's a nullable field you can do things like this:
> 
> `va.location = Location.objects.filter(name='whatever').first()`
> 
> And if there is not a location with name `whatever` then it will get assigned null.
> 
> Then you can assume any VA with a null location is "Unknown".
> 
> Seems like that could simplify a lot of logic - especially what's in `set_null_location` as you could just set `va.location = None`
> 
> But maybe I am missing something that prevents us from doing that.

Good point. I honestly went this route because there seemed to be consensus between Sarah and Pete that it was the right approach and I was building off of the work that they'd already done. But this would certainly simplify things. I'll experiment with removing the null constraint quickly and see if I can't get it to work.

timster (approved):

PR #85, opened 2021-05-04
fixed timeout issue preventing dashboard data from loading

babraham33:
This is really a bandaid but the dashboard should at least load now with default settings. We have some scaling issues that we'll need to address in future issues


ajbarnes (approved):
Pulled changes locally and was able to build without errors and successfully run tests. UI review revealed an age_group bug that was fixed after discussion. LGTM!

PR #87, opened 2021-05-06
Optimize load_va_data.

timster:
This optimizes load_va_data. Got it down from 20+ seconds to about 4 seconds on my machine.


babraham33:
Looks good! Just had to revert location_types back to a dict indexed by depth and sort by depth because my code assumes location types are sorted that way. Feel free to look it over once more before we merge

babraham33 (approved):

PR #88, opened 2021-05-07
converted all hidden divs to dcc stores

babraham33:
This PR introduces dcc.Stores as the primary mechanism to store client-side data for the dashboard instead of embedding it in the webpage's html. Now the raw data is no longer exposed in the user's DOM. Also fixed a few minor plotting bugs. 

Addresses https://www.pivotaltracker.com/n/projects/2455711/stories/177459626


timster:
I loaded the dashboard and it still works. I searched through the DOM and confirmed the data is not stored there anymore. Looks good.

timster (approved):

PR #89, opened 2021-05-11
Added Dashboard Viewers role

timster:
This takes care of the following story:

https://www.pivotaltracker.com/n/projects/2455711/stories/177588723

Adds a "Dashboard Viewers" role. This role can only see the dashboard and download data. The downloaded data will have PII stripped since this role does not have access to view PII. This role does not have access to the Data Management tab.

Also includes a fix for null locations in the downloaded data. I discovered that while testing this.


ajbarnes (approved):
I pulled locally and was able to run pytest with no issues. I also tested around in the UI with a Field Worker I changed the permissions of to Dashboard Viewer against a regular Field Worker to confirm and it looked good. You might consider adding a Dashboard Viewer user via the `seed_demo_users` script as an improvement.

timster:
I removed the Download Data permission from the Dashboard Viewers and added it as a separate option on the users page.

This allows for more granular control over who can download data. It's now possible to have a user who can view the Dashboard but not download data:

Capture 2021-06-08 at 10 28 40 AM

Also added logic to only show the download button if the user has access.

@babraham33 Can you give this a look when you have a chance?

babraham33:
@timster Looks good. I was able able to create a dashboard user that could download data, another one that couldn't, and verified that the download button only rendered when for the first user. I did have to blow away my database and re-initialize groups to get everything to work though (just running the initialize_groups command again didn't do the trick). Not a big deal, especially if it's in a production environment where everything would have to be rebuilt again, but just thought I'd point it out.

PR #90, opened 2021-05-17
Add ability to import ODK form by name or ID.

timster:
# ODK PR

`va_data_management/utils/odk.py`
- Added `form_id` and `form_name` parameters to methods.
- This will allow `get_odk_form` to lookup form by either ID or name instead of just returning the last form in the list returned by ODK.

`va_data_management/management/commands/import_from_odk.py`
- Added `form-id` and `form-name` options to command.
- This allows user to specify which form should be imported.
- Make sure user only specifies `project-id` xor `project-name` as specifying both could be confusing.
- Make sure user only specifies `form-id` xor `form-name` as specifying both could be confusing.
- Print to stderr instead of raising ValueError (this is more standard in a management command).

`va_data_management/tests/tests_odk.py`
- Organized tests by classes.
- Cleaned up some existing tests.
- Enhanced tests to account for new `form_name` and `form_id` attributes.
- Added tests for management command to test for error conditions. 

`README.md`
- Updated examples of `import_from_odk` command.

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,12 +262,15 @@ Clicking the "Run Coding Algorithms" will execute the coding algorithms in the b
 
 ## Importing From ODK
 
-You can use the `import_from_odk` management command to import records from an ODK server like so. You must specify either project-name or project-id to import:
+You can use the `import_from_odk` management command to import records from an ODK server like so. 
+You must specify either project-name or project-id and form-name or form-id to import:
 
 ```
-./manage.py import_from_odk --project-id=1234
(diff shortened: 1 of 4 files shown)


babraham33:
Code looks good. Can confirm that this with two separate odk forms for the same project. I was able to Import one batch of records made with first VA form and another made with a second form without a problem by specifying form id from command line.  LGTM!

babraham33 (approved):

PR #92, opened 2021-05-24
WIP: Track individual coding and import runs

timster:
For imports and coding algorithm runs, it will create a BatchOperation instance.

For each BatchOperation, it tracks associated verbal autopsies.

For an import BatchOperation, tracks all the VAs that are successfully imported.

For a coding BatchOperation, tracks all the VAs without cause that are sent to the coding algorithm (regardless of whether they are successfully coded).


timster:
Screenshot of the Batch Operations page:

Coding algoritm is running:

batch-0

Coding algoritm finished:

batch-1

pkmitre (dismissed):
This looks nice! One thing that would be helpful is, for any given coding run, being able to tie the assigned CauseOfDeath or CauseCodingIssue to the VerbalAutopsy for that run. I could see having separate CodingBatch and ImportBatch models, and adding relations to the CodingBatch model, though there may be other ways to handle it.

timster:
> This looks nice! One thing that would be helpful is, for any given coding run, being able to tie the assigned CauseOfDeath or CauseCodingIssue to the VerbalAutopsy for that run. I could see having separate CodingBatch and ImportBatch models, and adding relations to the CodingBatch model, though there may be other ways to handle it.

I think this should be possible already unless I'm misunderstanding what you're getting at.

Each BatchOperation has a list of VerbalAutopsy - and each VerbalAutopsy has a CauseOfDeath and CauseCodingIssue associated.

So you could get all the CauseOfDeath instances for a specific BatchOperation if you wanted:

```python
CauseOfDeath.objects.filter(verbalautopsy__batchoperation=123)
```

pkmitre:
What about the case where 1) a coding run fails for a particular VA with some errors, 2) the VA is corrected to address the errors, and 3) the VA is coded in a subsequent coding run. In that case there will be two coding runs associated with that particular VA and I don't think it's possible to distinguish which coding run had the errors and which had the successful coding result.

timster:
You're right. It doesn't handle that use case. Let me think over the design a bit and come up with something that addresses that.

pkmitre:
One thought:

An ImportBatch model that has a 1-to-many relationship with the VerbalAutopsy model

A CodingBatch model that has a 1-to-many relationship with both the CauseOfDeath and CauseCodingIssue models

The CodingBatch model then has an implicit 1-to-many relationship with the VerbalAutopsy model through both the CauseOfDeath and CauseCodingIssue models

timster:
Updated to split out `ImportBatch` and `CodingBatch` models. I think it's a better fit to track everything that happens for importing and coding.

In this test case below there were 39 VAs imported (as you can see by the import).

The first time coding algorithm ran, it coded 27 VAs and 12 had errors (there were also 195 warnings).

The second time coding algorithm ran, it tried to code the 12 that had errors (since they didn't have a COD) and they all failed again.

So it keeps a full history of all errors/warnings/successful across all coding runs.

Screen Shot 2021-06-09 at 3 24 06 PM

PR #93, opened 2021-06-02
removed ability to edit instanceid

babraham33:
Users can no longer edit instanceid from the GUI. Also made it an immutable field so can't be changed from the backend either once it's set.  Addresses https://www.pivotaltracker.com/n/projects/2455711/stories/177588477


ajbarnes (approved):
Looks like a good simple change. To confirm, does this close a pivotal issue? If so, could you also update that/mention it here for tracking? LGTM after bringing up to date with `main` 👍

PR #94, opened 2021-06-11
Bulk user create improvements

babraham33:
PR to improve bulk user creation capability. Bulk upload command now supports the same options currently available in user creation form but is flexible enough to handle fewer fields. Current customizable permissions include `view_pii` and `download_data`, but this code should pick up new permissions as they're added to system. This PR added/edited the following management commands: 
1. `get_user_form_template` - export fields being recorded in current user form. Admins can use this as a blueprint to structure their new user file.
2. `export_user_info` - export anonymized user info including all relevant permissions - handy for logging and pilot tracking
3. `bulk_load_users` - create users from a csv file

Additionally, added some tests to validate new user properties against the original data (see `users/tests/test_bulk_create.py`)

Addresses https://www.pivotaltracker.com/n/projects/2455711/stories/177793955


user_list.csv
```diff
@@ -0,0 +1,98 @@
...
+8a1c12d6-47aa-404f-bbb0-c4c69f39babc,True,False,Field Workers,Kaoma District Hospital
+7aa98c10-86fb-40d7-bb86-da06979d341d,True,False,Field Workers,Kaoma District Hospital
+aa40c270-6647-4502-ba10-b00074b7dafc,True,False,Field Workers,Lewanika General Hospital
+4f57804b-3fe7-4dd7-ae60-8f0a747c3dd4,True,False,Field Workers,Lewanika General Hospital
+967c897d-672a-4483-be97-168448a3c4c5,True,False,Field Workers,Senanga General Hospital
+74a67d2b-279f-480a-812f-5b3cf57aa15d,True,False,Field Workers,Senanga General Hospital
+548414fb-137c-4582-bf13-332142caa786,True,False,Field Workers,Mwandi Mission Hospital
+656ef15f-4df9-43df-9f2f-c34e88d7fd63,True,False,Field Workers,Mwandi Mission Hospital
+5451b5c5-e910-48cb-ad67-660c7bd7fbad,True,False,Data Viewers,
+f239287c-eb0e-4d03-b337-8ed5009bdd07,True,False,,
```
ajbarnes, line 98: Should this file be tracked?

va_explorer/users/forms.py
```diff
@@ -231,8 +233,12 @@ def save(self, commit=True):
...
             # https://github.com/pennersr/django-allauth/blob/c19a212c6ee786af1bb8bc1b07eb2aa8e2bf531b/allauth/account/utils.py
             # setup_user_email(self.request, user, [])
-
-            get_adapter().send_new_user_mail(self.request, user, password)
+            if email_confirmation:
+                get_adapter().send_new_user_mail(self.request, user, password)
+                
+            else:
+                print(f"="*20)
+                print(f"Created user with email {user.email} and temp. password {password}")
```
ajbarnes, line 241: Should there be a comment here mentioning how this is a work-around for the current email server blocking certain outbound email to certain providers and we'd like to eventually not log temp passwords?

ajbarnes (changes requested):
This looks good overall. I was able to pull and run the tests + new tests without issue. I left some comments during review to look over and would also like to request that the new management commands be added to the README (a short blurb to let at least let new admins know that they're available should be fine)

user_list.csv
babraham33: ah no, sorry forgot to add this to .gitignore

va_explorer/users/forms.py
babraham33: yes, definitely. I will expose this argument to the management command so choolwe can automatically send out emails when running it

PR #95, opened 2021-06-17
Volumize Django logfiles

ajbarnes:
Directs existing log configurations to use a single directory and uses that directory as a volume to make said logfiles available outside the container.


babraham33 (approved):

babraham33:
FYI I had to create the `va_logs/logs` folder manually to get the app and tests to run. (Fixed by adding this folder to the repo). However, while the volume mount worked properly (va_explorer_logs folder created in repo after docker image is run) the app failed write logs to file (no events.log or data_ingest.log). I got a test message to write out to `events.log` by sshing into the django container and manually calling the `write_va_log` with dummy text. So django should be recognizing the logger, but it fails to do so from within the app. Not sure why this is happening.

PR #97, opened 2021-06-24
added ability to specify subset of users for when exporting anonymized user info

babraham33:
added an optional` --user_file` argument to the `export_user_info` command allowing admin to specify subset of users to export info for. (txt file with user emails, one per line) When no user file passed, the command will export info on all users in the system.


user_list.csv
```diff
@@ -0,0 +1,5 @@
+uuid,is_active,is_staff,role,location_restrictions,view_pii,download_data
+55d8bfef-aa8e-42b8-b6c9-b56e95804378,True,False,,,NA/Unassigned,NA/Unassigned
+f10c95cd-4e3d-46e9-9c33-3c11f47d3bca,True,False,Field Workers,Kapiri General Hospital,NA/Unassigned,NA/Unassigned
+88ce9132-131e-4f4f-818a-93d8a47b0af9,True,False,Field Workers,,NA/Unassigned,NA/Unassigned
+057ff0cc-853b-4ab0-a4f7-ebb86c4ef8f9,True,False,Data Viewers,,NA/Unassigned,NA/Unassigned
```
ajbarnes, line 5: Was this file meant to be included in the MR? If so, would it belong at the root dir?

README.md
```diff
@@ -120,8 +120,9 @@ Once the prerequisites are available, VA Explorer can be installed and demonstra
     `./manage.py get_user_form_template --output_file <FILENAME>`
 
   * Export anynymous info for all users in system 
-    `./manage.py export_user_info --output_file <FILENAME>`
-      This will export an anymymous user ID, user roles, geographic restrictions and privileges to a csv file. Can be used to track user activity in logs withou compromising thier PII.
+    `./manage.py export_user_info --output_file <FILENAME> --user_file=<FILENAME>`
+      This will export anymymous user IDs, user roles, geographic restrictions and privileges to a csv file. Ultimately file can be used to track user activity in logs withou compromising thier PII. By default, itexports info on all users in the system, but you can choose to filter down to a select list of users by setting the --user_file argument to a .txt file with all user emails (one per line) you'd like to know about. In this case, the command will tell you which emails failed to match users in the database.
```
ajbarnes, line 124:
```suggestion
      This will export anonymous user IDs, user roles, geographic restrictions and privileges to a `.csv` file. Ultimately, the file can be used to track user activity in logs without compromising their PII. By default, it exports info on all users in the system, but you can choose to filter down to a select list of users by setting the `--user_file` argument to a `.txt` file with all user emails (one per line) you'd like to know about. In this case, the command will tell you which emails failed to match users in the database. 
```

ajbarnes (changes requested):
MR looks good overall. I left a comment and a suggestion, but think it's otherwise GTM. 👍

user_list.csv
babraham33: No - think I forgot to add it to `.gitignore`. Will do that now.

ajbarnes (approved):

PR #100, opened 2021-07-13
made download data optional for data viewers

babraham33:
also removed dashboard viewer role and updated pytests. To test: 
1. create a Data Viewer through the gui that cannot download data. 
2. login as that user, browse to dashboard, and confirm that no download data button is visible
3. sign out, login as admin, and update Data Viewer's permissions to be able to download data
4. login as Data Viewer again, browse to dashboard and confirm you can download data
5. run pytest and confirm the user tests run successfully

**Note** This may introduce an issue where Data Viewers created before the change can no longer download data (because the download permission is no longer tied to Data Viewer Role). If this happens, you can assign download-data permissions to all existing data_viewers by doing the following:  

open django shell - `./manage.py shell`

run the following code:

```
from va_explorer.users.models import User
data_viewers = User.objects.filter(groups__name='Data Viewers')
for user in data_viewers:
	if not user.can_download_data:
		user.can_download_data = True
		user.save()
```

Finally, added a few other bug fixes. These include
1. Changing data management tab to display/query for interviewer name instead of deceased name
2. Fixing user_create command to accept a sublist of users again
3. Ensured the default user group is Data Viewer when raw group doesn't match known groups during user creation
4. Upped the sensitivity of fuzzy matching for user creation form (was previously generating false positives)


ajbarnes (approved):
Code review looked good and I was able to sync with you offline to fix a bug/ ask questions - thank you. I pulled again and confirmed the changes + passing tests. LGTM 👍

PR #119, opened 2021-07-22
changed group and user ownership to django user/group to prevent buil…

babraham33:
…d issue


ajbarnes (approved):
Pulled locally and confirmed passing tests

PR #146, opened 2021-08-04
Add filters and role based name data to data management module

### LINKED ISSUE #133: Update Data Management page to display role-based name and support cause queries
As a user of the data management page, I expect to be able to query by cause of death.
As an admin, I expect to see names of both the interviewer and the deceased. As a data manager/data viewer I expect to see the name of the interviewer. As a field worker I expect to see the name of the deceased.

**What is the actual state?**
The data management page does not allow for queries by cause of death.
The data management page only shows the name of the interviewer for all roles.

**Relevant context**
- `/va_explorer/va_data_management/filters.py`
- `/va_explorer/va_data_management/views.py`

ajbarnes:
This PR adds:
- Ability to filter data mgmt by cause of death
- Ability to view, and filter by name of interviewer for non field worker roles
- Ability to view, and filter by name of deceased for all roles

Tangentially, it changes views so that field workers now have the ability to view VA warnings and errors

To test:
1. Login as a non field worker role and navigate to the data mgmt module
2. Observe filter form fields for cause of death, name of interviewer, and name of deceased
3. Attempt to filter by any or all of these properties and confirm expected returned VA results
4. Repeat step 1 for a field worker role user
5. Observe absense of interviewer from previously mentioned fields
6. Bonus: Confirm that querying a redacted (you as dev know what field contains but user doesn't) field doesn't return data for known (to you as dev) data 

Closes #133

### DIFF
--- a/va_explorer/templates/partials/_va_table.html
+++ b/va_explorer/templates/partials/_va_table.html
@@ -5,13 +5,14 @@
     <tr>
       <th>ID</th>
       <th>Date</th>
-      <th>Interviewer</th>
+      <th>Deceased</th>
+      {% if not user.is_fieldworker %}
+        <th>Interviewer</th>
+      {% endif %}
       <th>Facility</th>
       <th>Cause</th>
-      {% if not user.is_fieldworker %}
-        <th>Warnings</th>
-        <th>Errors</th>
-      {% endif %}
+      <th>Warnings</th>
+      <th>Errors</th>
       <th>View</th>
     </tr>
   </thead>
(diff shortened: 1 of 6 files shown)


babraham33:
Overall this looks great, but ran into a few minor issues. Can confirm that the filtering works as intended for non-fieldworkers. and that the Interviewer and COD fields are hidden from Field Workers. However, things went awry for me when I tried filtering as a FW. In this example, field_worker10@example.com is assigned to Nchanga North Central Hospital and has completed 10 VAs. When I load the initial data management page. Everything looks fine in the default view: 

Logged in as fieldworker10 who’s assigned to Nchanga North General Hospital and linked to 10 VAs
Verbal Autopsies

**The problem**: When I add a search param/try to filter data, it shows me _all_ Nchanga Hospital VAs, not solely his VAs. When I try filtering by Deceased (a redacted field) this is what I get:
Verbal Autopsies

Seems to be returning doing some filtering because the queryset is much larger now. Turns out that the first 6 results, all have Jane as the deceased, suggesting the filter still worked for the redacted field: 
from va_explorer  va_data_management models import VerbalAutopsy, Location

babraham33:
Can confirm that the issues mentioned above are fixed now. Confirmed that field workers can no longer filter by sensitive fields if they don't have PII. Also confirmed that granting them access to PII allows them to filter by sensitive fields. I think we'll eventually want to generalize the logic to exclude any filter fields in the PII_FIELDS list but definitely a problem for another day. For now, I say LGTM!

babraham33 (approved):

PR #147, opened 2021-08-05
location-ingest-improvements-v2

babraham33:
This PR changes location ingest logic to do soft inserts/updates instead of clearing the entire database by default (which causes a cascading effect). Also added a few minor optimizations and bug fixes around va data management and the dashboard. Finally, lays foundation for future POI plotting by adding optional lat/lon and address fields to Location model.


ajbarnes:
Closing in favor of #196

PR #172, opened 2021-09-24
fixed datetime issue preventing data from rendering in supervision tab

babraham33:
Was still unable to reproduce the error Chooolwe saw but added a patch that _should_ fix the it.


ajbarnes (approved):
Pulled locally, successfully ran test suite and visually confirmed expected behavior. Added an incidental change to fix a fieldworker check bug discovered during review.

PR #184, opened 2021-12-28
Issue102

HarunFeraidon:
replacing `dk` and `nan` in data_management view by using a Django template filter.
image

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -34,7 +34,8 @@ channels==2.4.0
 daphne==2.5.0
 
 # Plotly
-scipy==1.5.2
+scipy==1.5.2; python_version == '3.7'
+scipy==1.7.3; python_version > '3.7'
 
 # Dash
 dash==1.14.0
--- a/va_explorer/templatetags/va_explorer_tags.py
+++ b/va_explorer/templatetags/va_explorer_tags.py
@@ -2,6 +2,8 @@ import re
 
 from django import template
 from django.urls import NoReverseMatch, reverse
+from django.template.defaultfilters import stringfilter
+
 
 from urllib.parse import urlencode
 from collections import OrderedDict
@@ -11,6 +13,18 @@ from va_explorer.va_data_management.models import REDACTED_STRING
 
 register = template.Library()
 
+@register.filter
+def replace(value):
+    if("dk" in value):
+        return "Don't know"
+    elif("nan" in value):
+        return "Not a number"
+    else:
+        return value
+
+@register.filter
+def upper(value):
+    return value.upper()
 
 @register.simple_tag(takes_context=True)
 def active(context, pattern_or_url):
--- a/va_explorer/templates/va_data_management/show.html
+++ b/va_explorer/templates/va_data_management/show.html
@@ -43,7 +43,9 @@
             <tr>
               <td><b>{{ field.name }}</b></td>
               <td>{{ field.label }}</td>
-              <td>{% pii_filter field.name field.value %}</td>
+              <td>
+                {% pii_filter field.name field.value|replace %}
+              </td>
             </tr>
           {% endfor %}
         </tbody>


requirements/base.txt
```diff
@@ -34,7 +34,8 @@ channels==2.4.0
 daphne==2.5.0
 
 # Plotly
-scipy==1.5.2
+scipy==1.5.2; python_version == '3.7'
+scipy==1.7.3; python_version > '3.7'
 
 # Dash
```
deastman, line 40: This is covered in your other PR, so can be removed from this one. We would just merge your other PR first.

va_explorer/templatetags/va_explorer_tags.py
```diff
@@ -11,6 +12,27 @@
...
+    value = value.strip()
+    if("dk" == value or "DK" == value):
+        return "Don't know"
+    elif("nan" == value):
+        return "n/a"
+    elif("veryl" == value):
+        return "Very Low"
+    elif("ref" == value):
+        return "refuse to answer"
+    else:
```
deastman, line 26: Can you make the capitalization consistent with the other replacements, i.e., "Refuse to Answer"

va_explorer/templatetags/va_explorer_tags.py
```diff
@@ -11,6 +12,27 @@
...
+        return value
+
+@register.filter
+def upper(value):
+    return value.upper()
+
+@register.filter
+def capitalize_every_first_letter(value):
+    return ' '.join(s[:1].upper() + s[1:] for s in value.split(' '))
```
deastman, line 36: Are these filters used somewhere?

va_explorer/templatetags/va_explorer_tags.py
```diff
@@ -11,6 +12,27 @@
 
 register = template.Library()
 
+@register.filter
+def replace(value):
+    value = value.strip()
```
deastman, line 17: I think you can remove the parenthesis with these conditionals: https://google.github.io/styleguide/pyguide.html#33-parentheses

va_explorer/templatetags/va_explorer_tags.py
HarunFeraidon: They're not, I liked the idea of including some filter functions to serve as an example of how they are written, or also in the case that we want to come back to use them again.

Also, per .title(), it doesn't work for cases of strings with apostrophe `'`, such as somebody mentioned in [this thread](https://stackoverflow.com/questions/8199966/python-title-with-apostrophes)


I can take these functions out if preferred

va_explorer/templatetags/va_explorer_tags.py
HarunFeraidon: do you mean remove the parentheses such that its `value.strip` and not `value.strip()` ?

deastman, line 17: I might just lowercase `value` and then check on the lowercased version in your conditionals. Then you would catch, "DK", "dk", "Dk", "NaN", "NAN", "REF", "ref" etc.

deastman:
I noticed that the questionnaire also uses snake case in various places, e.g.,

local_form_of_tobacco
lower_abdomen
relative_friend

Is it in scope of this issue to convert these to a more readable format, like title case? If not, no problem.

va_explorer/templatetags/va_explorer_tags.py
deastman: No, sorry for being unclear. I mean `if("dk" == value or "DK" == value)` --> `if "dk" == value or "DK" == value`

This is purely stylistic.

HarunFeraidon: I don't think that would work because this would lowercase all value's, because my `else` condition is to just return whatever input was received.

va_explorer/templatetags/va_explorer_tags.py
deastman: I see. I think this custom method needs to be tested a bit more and revised. 

E.g., for string "my cAT", the custom method returns 'My CAT', which I don't think is quite correct. I think the intention is to use titlecase and return 'My Cat'. 

"My cAT".title() returns 'My Cat'. 

If you want to leave `capitalize_every_first_letter` in the code after debugging it, that seems fine. I don't think it's worth having a filter for upper, since it's just wrapping an existing Python function.

va_explorer/templatetags/va_explorer_tags.py
deastman: `lower_value = value.lower() `and use `lower_value` that in your conditionals, then return the original `value`.

va_explorer/templatetags/va_explorer_tags.py
HarunFeraidon: I'll just remove the two functions from the PR since I think they're not used anyway. 
I intended on using them at first but later decided not to but still kept them in because I thought they were helpful to look at when writing new filters.

HarunFeraidon:
> I noticed that the questionnaire also uses snake case in various places, e.g.,
> 
> local_form_of_tobacco lower_abdomen relative_friend
> 
> Is it in scope of this issue to convert these to a more readable format, like title case? If not, no problem.

I think it's possible, one concern I have is with the field that is 'username', some users can have snake case or any instance of '_' in their username, which would get filtered.
I can make another ticket for it and come back to finding a solution for it later where the filter would apply to all fields except for the on of 'username'

deastman (approved):
Looks great! 🎉

PR #195, opened 2022-01-12
Add docker-integrated cronjobs for ODK Import via ofelia 

### LINKED ISSUE #106: Automate data import from ODK
**Is your feature request related to a problem? Please describe.**
As an admin I would like data imports from ODK to happen automatically without having to exec into the VA Explorer docker container.

**Describe the solution you'd like**
 A good solution might be to use something like [ofelia](https://github.com/mcuadros/ofelia) to add configurable cron-like functionality that integrates well with docker-compose and current installation steps

**Describe alternatives you've considered**
- cron within the docker container (not recommended because cron could cause overlap if/when containers are scaled)
- [schedule](https://pypi.org/project/schedule/) (same as above but built-in to python)
- cron on the host system (not recommended because it adds cron installation step to VA Explorer setup on new systems)

ajbarnes:
This issue introduces cron style automation to VA Explorer, starting with ODK imports so that admins do not have to manually run this management command.

Specifically it:
- Updates docker-compose to add [ofelia](https://github.com/mcuadros/ofelia) service specifically for docker environments that avoids the potential problems discussed in alternative consideration #106 lists
- Adds an initial cronjob to our django container to run the `import_from_odk` management command every night at 3am
- Updates django service to ensure all ODK specific variables are injected into the container from `.env` at build time
- Fixes an issue in `odk.py` where True/False in `.env` would be parsed as a string and cause requests.get to throw errors.

To verify this was working as expected I:
- ensured that my `.env` was populated with `ODK_HOST`, `ODK_SSL_VERIFY`, `ODK_PROJECT_ID`, `ODK_FORM_ID`, `ODK_EMAIL`, and `ODK_PASSWORD` with email and password being auth info for a local instance of the odk-central application
- temporarily changed to this line in docker-compose
```
ofelia.job-exec.odk-import.schedule: "@every 1m"
```
- re-ran `docker-compose up -d --build`
- observed the job successfully ("failed: false, ..., error: none") running every minute via `docker-compose logs -f django ofelia`

**Note:** currently it seems that having these variables defined in `.env` is not compatible with running pytest as it relies on default values. It may be worth opening an issue to address this, in the meantime I commented out my `.env` before running `pytest`

Closes #106

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -332,5 +332,6 @@ va_explorer/media/
 .env
 .envs/*
 !.envs/.local/
+*.pem
 
 pgdata
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,6 +5,14 @@ volumes:
   production_postgres_data_backups: {}
   
 services:
+  ofelia:
+    image: mcuadros/ofelia:latest
+    depends_on:
+      - django
+    command: daemon --docker
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock:ro
+
   pycrossva:
     image: va_explorer/pycrossva
     build: https://github.com/VA-Explorer/pyCrossVA.git#microservice-experiment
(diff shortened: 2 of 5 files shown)


ajbarnes:
Closed as OBE

PR #196, opened 2022-01-14
Fix default behavior in location ingest command which causes cascading deletes

ajbarnes:
Revival of #135/#145 cherry-picking for current needs/ latest code.

Changes load_locations command to do inserts/updates instead of clearing all locations by default and prevents default behavior which caused a cascading effect on location delete. (Previous behavior can still be achieved by setting --delete-previous to True). To test the command's running properly:

1. reset db (running all necessary migrations, account creation, etc.)
2. run load_locations with old Zambia file. Confirm that all locations are created correctly
3. Load some VAs into system
4. run load_locations again with new Zambia file
5. Confirm that most locations were updated but that some new ones were added.
6. Confirm that all VAs are still in system (i.e. not deleted from Cascade)


babraham33 (approved):
Tested and confirmed the initial set of VAs do not get overwritten when adding new locations. Code looks good. LGTM!

PR #198, opened 2022-01-19
Cod config

babraham33:
This PR does three things: 
1. moves algorithm settings to .env file (defaulting to hardcoded values when not set in environment)
2. adds logic to run_coding_algorithms enabling you to re-code all VAs by overriding existing CODs (set `--overwrite=True`)
3. when `overwrite=True`, exports old CODs to a csv for safekeeping

To test:
1. run coding algorithms normally - `./manage.py run_coding_algorithms`
2. Change algorithm settings in env file (e.g. `INTERVA_MALARIA=l`)
3. re-run coding algorithms command with override=True - `./manage.py run_coding_algorithms --overwrite True`
4. Confirm old CODs are saved to csv
5. (optional) compare old CODs to new CODs and check if settings made a difference


va_explorer/va_data_management/management/commands/run_coding_algorithms.py
```diff
@@ -1,15 +1,47 @@
+import argparse
+import pandas as pd
+import time
 from django.core.management.base import BaseCommand
+from django.db.models import F
 from va_explorer.va_data_management.utils.coding import run_coding_algorithms
+from va_explorer.va_data_management.models import VerbalAutopsy, CauseOfDeath
```
deastman, line 9: @babraham33 Can you add the algorithm settings here and first validate that the values the user has set in the .env file are valid? If the user-supplied values in the .env are invalid, either fallback to the default setting or exit the program. 

Example: In the .env file, I set `INTERVA_HIV=t` (invalid). But, that wasn't hit until the cause of death table had already been cleared. We probably want the program to exit if an invalid setting is passed in before clearing the previous causes of death.

va_explorer/va_data_management/management/commands/run_coding_algorithms.py
```diff
@@ -1,15 +1,47 @@
...
+        print('clearing old CODs...')
+        cod_fname = kwargs.get('cod_fname', 'old_cod_mapping.csv')
+        # export original CODs (and corersponding VA IDs) to flat file
+        va_cods = VerbalAutopsy.objects.only('id', 'causes').select_related('causes').values(va_id=F('id'), cause=F('causes__cause'))
+        va_cod_df = pd.DataFrame(va_cods).dropna()
+        # only export if VAs have been coded
+        if not va_cod_df.empty:
+            va_cod_df.to_csv(cod_fname, index=False)
+            print(f'exported {va_cod_df.shape[0]} old CODs to {cod_fname}')
+        else:
```
deastman, line 43: I wonder if we should just export the entire COD table to csv, as is. Use case: if I keyboard interrupt 

`./manage.py run_coding_algorithms --overwrite True`

I might want to re-import the entire COD table into the db.

va_explorer/va_data_management/management/commands/run_coding_algorithms.py
```diff
@@ -1,15 +1,47 @@
...
         results = run_coding_algorithms()
         num_coded = len(results['causes'])
         num_total = len(results['verbal_autopsies'])
         num_issues = len(results['issues'])
-
+        self.stdout.write(f'DONE. Total time: {time.time() - ti} secs')
         self.stdout.write(f'Coded {num_coded} verbal autopsies (out of {num_total}) [{num_issues} issues]')
+
+    def clear_and_save_old_cods(self, *args, **kwargs):
+
```
deastman, line 33: Don't need *args here. I would also just pass `cod_fname` directly to `clear_and_save_old_cods` since that's the only option that's being used in this function. 

So in `handle` on line 22 you could just do (since you already have the fallback option set in parser):
` self.clear_and_save_old_cods(options['cod_fname'])`

va_explorer/va_data_management/management/commands/run_coding_algorithms.py
```diff
@@ -1,15 +1,47 @@
...
 class Command(BaseCommand):
 
     help = 'Run cause coding algorithms'
 
+    def add_arguments(self, parser):
+        # whether or not to overwrite existing CODs - if True, will clear all CODs before running
+        parser.add_argument('--overwrite', type=bool, nargs='?', default=False)
+        parser.add_argument('--cod_fname', type=str, nargs='?', default='old_cod_mapping.csv')
+
     def handle(self, *args, **options):
```
deastman, line 19: Don't need *args

deastman:
Works great, @babraham33! Couple of minor suggestions. Also, will this be added to the README?

va_explorer/va_data_management/management/commands/run_coding_algorithms.py
babraham33: Great point. Will do!

va_explorer/va_data_management/management/commands/run_coding_algorithms.py
```diff
@@ -1,15 +1,52 @@
...
+            results = run_coding_algorithms()
+            num_coded = len(results['causes'])
+            num_total = len(results['verbal_autopsies'])
+            num_issues = len(results['issues'])
+            self.stdout.write(f'DONE. Total time: {time.time() - ti} secs')
+            self.stdout.write(f'Coded {num_coded} verbal autopsies (out of {num_total}) [{num_issues} issues]')
+        else:
+            raise ValueError(f"At least one invalid algorithm setting in: \n {ALGORITHM_SETTINGS}. See va_data_management.utils.coding.py for valid setting values." )
+
+    def clear_and_save_old_cods(self, cod_fname=None):
```
deastman, line 37: If you just pass in cod_fname (without setting it to none), it will always be populated: either with the custom file name that was passed in via the CLI or with 'old_cod_mapping.csv', since that was the default in` parser.add_argument()`

va_explorer/va_data_management/management/commands/run_coding_algorithms.py
```diff
@@ -1,15 +1,52 @@
...
+        
+        print('clearing old CODs...')
+        if not cod_fname:
```
deastman, line 40: Don't need to do this check because it will always be either the custom file name or 'old_cod_mapping.csv' because it's the default in `parser.add_argument()`

va_explorer/va_data_management/management/commands/run_coding_algorithms.py
```diff
@@ -1,15 +1,52 @@
...
+        # whether or not to overwrite existing CODs - if True, will clear all CODs before running
+        parser.add_argument('--overwrite', type=bool, nargs='?', default=False)
+        parser.add_argument('--cod_fname', type=str, nargs='?', default='old_cod_mapping.csv')
 
-        self.stdout.write(f'Coded {num_coded} verbal autopsies (out of {num_total}) [{num_issues} issues]')
+    def handle(self, **options):
+        ti = time.time()
+        # validate algorithm settings first. Only proceed if settings are valid. 
+        settings_valid = validate_algorithm_settings(ALGORITHM_SETTINGS)
+        if settings_valid:
```
deastman, line 23: Probably don't need `settings_valid` variable since it's only being using in line 23. 

Can just do:
`if validate_algorithm_settings(ALGORITHM_SETTINGS)`:

va_explorer/va_data_management/utils/coding.py
```diff
@@ -12,13 +12,51 @@
...
+
+# validates provided interva settings against interva5 param value sets
+def validate_algorithm_settings(settings=ALGORITHM_SETTINGS):
+    # TODO: make this optional parameter once we support other algorithms
+    algo_key = "INTERVA"
+    param_opts = ALGORITHM_PARAM_OPTIONS[algo_key]
+    setting_keys = set(settings.keys())
+    common_keys = setting_keys.intersection(param_opts.keys())
+    
+    if len(common_keys) != len(setting_keys):
```
deastman, line 49: Is this for when we support additional algorithms and the `algo_key` is passed in? If so, can we comment out for now and add another # TODO?

babraham33: Yeah exactly. Sure thing

va_explorer/va_data_management/management/commands/run_coding_algorithms.py
```diff
@@ -1,15 +1,52 @@
...
+
+            print('coding all eligible VAs... ')
+            results = run_coding_algorithms()
+            num_coded = len(results['causes'])
+            num_total = len(results['verbal_autopsies'])
+            num_issues = len(results['issues'])
+            self.stdout.write(f'DONE. Total time: {time.time() - ti} secs')
+            self.stdout.write(f'Coded {num_coded} verbal autopsies (out of {num_total}) [{num_issues} issues]')
+        else:
+            raise ValueError(f"At least one invalid algorithm setting in: \n {ALGORITHM_SETTINGS}. See va_data_management.utils.coding.py for valid setting values." )
```
deastman, line 35: With the stack trace that's shown due to the exception, it's hard to read your error message. Maybe just:

```
            print(f"At least one invalid algorithm setting in: \n {ALGORITHM_SETTINGS}. See va_data_management.utils.coding.py for valid setting values. \n"
                  f"Exiting...")
            exit()
```

va_explorer/va_data_management/utils/coding.py
```diff
@@ -12,13 +12,51 @@
...
+    # describes prevalence of HIV. Should be "h"(high),"l"(low), or "v"(very low).
+    'HIV': os.environ.get('INTERVA_HIV', 'h'),
+    # describes prevalence of Malaria. Should be "h"(high),"l"(low), or "v"(very low).
+    'Malaria': os.environ.get('INTERVA_MALARIA', 'l'),
+    # Whether to include group code in the output causes
+    'groupcode': os.environ.get('INTERVA_GROUPCODE', 'True')
+}
+
+# validates provided interva settings against interva5 param value sets
+def validate_algorithm_settings(settings=ALGORITHM_SETTINGS):
```
deastman, line 42: Is settings being passed in for future work? Otherwise, `ALGORITHM_SETTINGS` is in the global context and can be used in the method as is.

babraham33: Not sure. Wanted to make the function flexible but don't think it people would use would use it in a one-off fashion. Will change to just use ALGORITHM_SETTINGS

va_explorer/va_data_management/management/commands/run_coding_algorithms.py
```diff
@@ -1,15 +1,50 @@
...
+    def add_arguments(self, parser):
+        # whether or not to overwrite existing CODs - if True, will clear all CODs before running
+        parser.add_argument('--overwrite', type=bool, nargs='?', default=False)
+        parser.add_argument('--cod_fname', type=str, nargs='?', default='old_cod_mapping.csv')
 
-        self.stdout.write(f'Coded {num_coded} verbal autopsies (out of {num_total}) [{num_issues} issues]')
+    def handle(self, **options):
+        ti = time.time()
+        # validate algorithm settings first. Only proceed if settings are valid. 
+        if validate_algorithm_settings(ALGORITHM_SETTINGS):
```
deastman, line 22: Remove `ALGORITHM_SETTINGS` since validate_algorithm_settings takes no params now

README.md
```diff
@@ -305,7 +308,19 @@ This will list a brief report of results in the following format:
 Coded 24 verbal autopsies (out of 30) [6 issues]
 ```
 
-You will receive an error message if pyCrossVA or InterVA5 are unavailable.
+If you'd like to re-code all existing VAs in the system (overwriting previous COD assignments) you can do so by passing `overwrite True` to the command-line call. This will save a backup table of old COD assignments, clear all CODs in the database, and re-run the coding algorithm on all eligible VAs.
+
```
deastman, line 312: By default, old COD assignments will be saved to a .csv file named "old_cod_mapping.csv." You can customize the name of this file by passing `--cod_fname <filename>` as an optional parameter.

deastman:
@babraham33 One small thing that needs to be changed that's causing an error and one potential addition to the README. This is great!! 🎉

deastman (approved):
🚀

PR #199, opened 2022-01-19
Issue140 - added ability to filter through IDs in data management tab

HarunFeraidon:
This PR closes issue 140.
Changes: The changes add ability to filter with ID in the data management view. Currently, you can only filter with 1 ID in the field.
Bad inputs are validated.

These changes are useful as they will add another way to filter for rows of VAs.

Testing instructions:
1.) Open data_management tab with mock data installed.
2.) Try filtering and rows based on ID value. You can also check your mock data file to make sure ID is accurate. Make sure the view updates the rows of VAs for the filter items.
3.) For validation, try entering bad input. Such as "123abc" or "12$" or "5b", you can also enter negative numbers such as "-4" which are also validated out.


va_explorer/templates/partials/_va_table.html
```diff
@@ -3,6 +3,7 @@
 <table class="table table-sm">
   <thead>
     <tr>
+      <th><a href="?{% sort_url 'Id10010' direction='-' %}">ID</a></th>
```
ajbarnes, line 6:
```suggestion
      <th><a href="?{% sort_url 'id' direction='-' %}">ID</a></th>
```
I think this is the actual desired behavior here

ajbarnes (changes requested):
Pulled locally confirmed testing and followed included instructions without issue. I highlighted one inconsistency in the column sorting found during code review and local pull. Once that's changed I say LGTM!

va_explorer/templates/partials/_va_table.html
HarunFeraidon: fixed, I copy-pasted from a line below and forgot to edit

ajbarnes (approved):

PR #209, opened 2022-02-15
Issue 190: Callback conversions to clientside

### LINKED ISSUE #190: Consider ways to prevent unnecessary API calls via clientside callbacks
Clientside callbacks are utilized in the context described in #173

**What is the actual state?**
No optimization

**Relevant context**
`va_explorer/va_analytics/dash_apps/va_dashboard.py`

Tyler-Manderfield:
Closes #190 

## Motivation
This PR attempts to speed up the loading of the dashboard by utilizing client-side callbacks instead of server-side callbacks when appropriate. The dashboard loading is not instantaneous and more data will only increase the amount of time it will take to load the data into the tool.

## Experiments & Results
To see whether client-side callbacks can actually speed up time, a few experiments were done. The first experiment examined the difference between the original server-side callbacks and the version of the code where the easiest callbacks were converted to client-side versions (for example, one of the callbacks just mapped the names of the tabs from something like "tab-1" to "COD Analysis"). There were four callbacks in total that were converted to client-side callbacks in this first experiment: 
- Dashboard Reset Callback (old function name: reset_dashboard) 
- Location Search Options (old function name: log_tab_value)
- Map metrics (old function name: get_metrics)
- Geographic Levels/View options (old function name: update_view_options)

The second experiment that was done examined the benefit of using Lodash instead of looping when we are working inside of the client-side callbacks (which are written in javascript). The use of Lodash did seem to slightly improve speed performance beyond using just loops. Compared to the old all-server-side version, there was a 8.33% improvement when using incorporating the client-side callbacks alone and an 8.56% improvement when using the client-side callbacks with Lodash. 

The final experiment looked at converting the callout boxes callback, which turned out to be a bit more complicated to convert due to having a few more connections to items on the server-side (namely the use of some functions to create the html objects that would display the callout boxes as well as references to global variables). After taking the time to successfully convert this callback to do more computations on the client-side, the speed of loading the dashboard decreased compared the previous version that utilized Lodash. The experiment found, across 20 page refreshes, that the all-server-side code took an average of 7.34 seconds to load the page, the Lodash with the preliminary client-side conversions took an average of 6.72 seconds to load the page, while the version that incorporated this final client-side converted callback took an average of 7.03 seconds. (Note that these times were recorded when the development toolbar was activated, which slows things down a little bit). The results indicate that converting the callout boxes callback to be a client-side callback increased the average time to load the page compared to leaving it as a server-side only callback. 

## Other miscellaneous changes made when addressing this issue
- Bug fix relating to the Trend Plot of the VA Trends tab of the Dashboard: it should now update its data and title when you select a new item from the Cause of Death dropdown towards the top of the page (Note: this is _not_ referring to the "Search COD Keywords" search bar but rather dropdown next to the Time Period dropdown and Reset button)
- Uncommented a line in plotting.py that was causing an issue with loading times when changing back and forth between different items in the Time Period dropdown. Previously, you could quickly change the Time Period and the dashboard would quickly load the new data, but changing it a second time would cause long waiting times.

## Testing Instructions
- The dashboard functionality should be exactly as it was before
- There should be an improvement in the speed of loading the new version of va_dashboard.py compared to the previous version.
    - In Microsoft Edge, you can check this by right-clicking your webpage, clicking "Inspect Page" and then finding "Network" in the tabs at the very top of the popout (you may have to check under  >> to see additional options). If you refresh the page you should be able to find how long it took about halfway down the page in the format of: "Finish : X.XX s".
- The data and title of the plot in the "VA Trends" Tab should update to when selecting a new item in the Cause of Death dropdown
    - Change the selection of the Cause of Death dropdown and note that the data and title changed. The title should no longer say "**All Causes** by Month" but "<selected COD here> by Month"
- There should no longer be a lag when switching between items in the Time Period dropdown:
    - Try changing the Time Period a few times and making sure that the data in the map on the left and plots on the right are updating in a reasonable time span (a few seconds). You shouldn't see 30 second+ wait times when changing the dropdown the second time.


va_explorer/va_analytics/dash_apps/va_dashboard.py
```diff
@@ -196,6 +198,7 @@
                                                 dcc.Dropdown(
                                                     id="view_level",
                                                     value="",
+                                                    # value=INITIAL_GRANULARITY,
```
HarunFeraidon, line 201: delete comment if not helpful

va_explorer/va_analytics/dash_apps/va_dashboard.py
```diff
@@ -812,34 +884,118 @@ def shift_granularity(current_granularity, levels, move_up=False):
 
 # =========Map Metrics =======================#
 # Top metrics to track for map dropdown
-@app.callback(
+# callback 5
```
HarunFeraidon, line 887: delete comment chunk if not helpful, especially if its code you commented out. If they're helpful comments keep them with good judgement

va_explorer/va_analytics/dash_apps/va_dashboard.py
```diff
@@ -812,34 +884,118 @@ def shift_granularity(current_granularity, levels, move_up=False):
...
+#                     .head(N)
+#                     .index.tolist()
+#                 )
+#     print("get metrics Callback", time.time() - START_TIME)
+#     return [{"label": LOOKUP["display_names"].get(m, m), "value": m} for m in metrics]
+
+### client-side conversion for get_metrics ###
+app.clientside_callback(
+    """
+    function(va_data, filter_dict, display_names){
```
HarunFeraidon, line 920: this function is really hard to read, its formatted as a string block but much of it is comments. please delete the commented out code especially if its unimportant

va_explorer/va_analytics/dash_apps/va_dashboard.py
```diff
@@ -851,32 +1007,77 @@ def get_metric_display_names(map_metrics):
...
+        
+            if(disable === false){
+                view_options = location_types;
+            }
+            else{
+                view_options = [];
+            }
+            options = [];
+            for(var i =0; i<view_options.length;i++){
+                options.push({"label":view_options[i].charAt(0).toUpperCase() + view_options[i].slice(1), "value":view_options[i]})
```
HarunFeraidon, line 1066: this is a personal preference, but I would declare `view_options[i].charAt(0).toUpperCase() + view_options[i].slice(1)` to a helpful variable name, since its a bit long of a function

va_explorer/va_analytics/dash_apps/va_dashboard.py
```diff
@@ -851,32 +1007,77 @@ def get_metric_display_names(map_metrics):
...
+            } 
+            return (options)
+    }
+    """,
+    Output(component_id="view_level", component_property="options"),
+        # Output(component_id="view_level", component_property="disabled"),
+        # Input(component_id="filter_dict", component_property="data"),
```
HarunFeraidon, line 1073: delete

va_explorer/va_analytics/dash_apps/va_dashboard.py
```diff
@@ -1410,8 +1777,10 @@ def trend_plot(va_data, timeframe, group_period, filter_dict=None, factor="All",
             figure = plotting.va_trend_plot(
                 plot_data, group_period, factor, title=plot_title, search_term_ids=search_term_ids, height=560
             )
+            # print("TIMEEEEEEEEEEEEEEEEEEE")
+            # print(time.time() - START_TIME)
+    print("trend plot Callback", time.time() - START_TIME)
```
HarunFeraidon, line 1782: nit but capitalization is off

va_explorer/va_analytics/dash_apps/va_dashboard.py
```diff
@@ -1410,8 +1777,10 @@ def trend_plot(va_data, timeframe, group_period, filter_dict=None, factor="All",
...
     return dcc.Graph(id="trend_plot", figure=figure, config=LOOKUP["chart_config"])
 
-
 # uncomment this if running as Dash app (as opposed to DjangoDash app)
-# app.run_server(debug= True)
+# app.run_server(debug= True)
```
HarunFeraidon, line 1786: delete...? I notice you added line 1786, but line 1785 was already there. So what was 1785 referring to before?

delete both imo

HarunFeraidon (changes requested):
first iteration, I haven't dived deep into the code yet

va_explorer/va_analytics/dash_apps/va_dashboard.py
Tyler-Manderfield: This was here before I started editing. Not sure why it looks like I edited it. Will leave it in in case Brendan wants to continue using it.

HarunFeraidon (approved):
lgtm

va_explorer/va_analytics/dash_apps/va_dashboard.py
```diff
@@ -1282,6 +1640,7 @@ def make_card(
 
 
 # =========Demographic plot logic==============================================#
+# callback 9
```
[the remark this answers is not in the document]
Tyler-Manderfield: Was just helpful so I could remember which ones I had converted vs did not. But was more for myself so they are deleted.

va_explorer/va_analytics/dash_apps/va_dashboard.py
Tyler-Manderfield: Good call. Updated.

PR #220, opened 2022-03-25
Bump versions for all possible packages except Django

ajbarnes:


deastman (commented):
@ajbarnes 

This looks good! I do have one question about the bump of `django-debug-toolbar`. As you had mentioned in Slack, now every time we use the manage.py facility a warning is printed out that referes django-debug-toolbar. E.g.,: 

```
?: (debug_toolbar.W006) At least one DjangoTemplates TEMPLATES configuration needs to have APP_DIRS set to True.
	HINT: Use APP_DIRS=True for at least one django.template.backends.django.DjangoTemplates backend configuration.
```
It might be unclear to someone else running these commands what the error refers to. It looks like this issue will be addressed in an upcoming release of django-debug-toolbar. See [here](https://github.com/jazzband/django-debug-toolbar/pull/1556). 

Is there any downside to going back to `3.1.1`?

ajbarnes:
Seems reasonable, I don't mind. Resolved in 3128e45db13d06bd2efa764df51dac7d2770149d

deastman (approved):

PR #234, opened 2022-04-25
Update va export to handle slow network connections

deastman:
# Description
This pull request aims to allow users with a slow network connection to download large csv and json files from the Export page. 

**Summary of Approach**
- In the export Django view, compress the csv and json files to a ZIP. 
- The zipping process takes a significant amount of time for large file sizes. If the user sees no visual response in the UI, they may think the download process has not been initiated and launch the download again. To mitigate this, we send the post via `XMLHttpRequest` so that we can display a modal via JS before we send the request. The modal notes that the download is being prepared and may take several minutes for large file sizes. The modal can be manually closed by the user, but will automatically close when the response returns back from the server. 

**Other optimizations include:** 
- Using vendor minified files when available for JavaScript and CSS libraries to speed up page load
- Moving some vendor files out of the base template and including only on pages where they are used

## Important Changes
Please list important files (meaning substantial or integral to the PR) along with a list of the general changes that should be highlighted for reviewers.

`.pre-commit-config.yaml`
- Bumps black version so the linting process runs on the CI

`va_explorer/dhis_manager/dhis.py`
- Fixes linting issue unrelated to the PR so that linting passes

`va_explorer/static/js/va_export.js`
- Shows the submit button when the page is completely loaded (this is required because we are posting the form via JS, so we need to ensure that this file loads before the user submits the form)
- Posts the form using JS and displays a modal to the user upon form submission
- Hides the modal when the request returns from the server
- Downloads the file 
- Shows an error modal if the server returns a status other than 200

`va_explorer/templates/base.html`
- Removes some vendor files from the base template to speed up page load (these files will be included w/in the individual pages that use them)

`va_explorer/templates/va_export/index.html`
- Adds bootstrap modals that appear upon Export submit and download error
- Hides the submit button (unhidden when `va_explorer/static/js/va_export.js` loads)

`va_explorer/va_analytics/dash_apps/va_dashboard.py`
- Fixes linting issue unrelated to the PR so that linting passes

`va_explorer/va_export/tests/test_views.py`
- Updates tests due to change in form action and download type (all downloads are zip files)

`va_explorer/va_export/views.py`
- Compresses the csv and json files to ZIPs 
- Uses POST method instead of GET 

## Testing Recommendations

Note: When repeatedly testing the download of large file sizes, be sure to close the Chrome DevTools pane. When I tried to test the largest sample data, i.e., the 100k dataset, and had the Chrome DevTools pane open, I encountered the error, `failed to load response data request content was evicted from inspector cache`.  See for more information: https://stackoverflow.com/questions/70008208/failed-to-load-response-data-request-content-was-evicted-from-inspector-cache

- Use Chrome DevTools to change the throttling profile in your browser. (Note: The slowest profile I tested with is the "Fast 3G" profile.) 
- Test the Export page for both json and csv formats on large datasets (e.g., 100k VAs)
- Confirm that a modal is shown when the form is submitted. Check the language of the modal to see if it communicates clearly what is happening. 
- Confirm that the download succeeds
- Unzip the file and confirm that the contents are valid 
- Ensure that all form filters still function correctly
- Confirm that the changes made due to linting errors (that are unrelated to this PR) do not change the application's functionality


babraham33 (approved):
Confirmed PR works as expected - was able to download 100k samples on fast 3g in both csv and json format. Also confirmed that pre-download filters still work with change. Code looks great - no qualms there. Ideally the modal would have a spinner or progress bar but that's a problem for another day. Functionally, this is ready to go. LGTM!

PR #238, opened 2022-06-06
hotfix: rm dangling staticfiles refs

ajbarnes:
- fix: rm old refs to .js/.css in favor of .min.js/.min.css + old files

Fixes error present in docker build where collectstatic is unable to find `bootstrap.js`, leading to cascading errors.


deastman (approved):
🎉

PR #249, opened 2022-07-19
Dashboard v2.0

haroutboujakjian:
This PR designs and deploys a new dashboard that is lighter, faster, and less clunky than the previous one. The old Plotly based dashboard is removed and replaced with a Vue based dashboard.

Important points:
- The new dashboard is built with Vue and [Vuesalize](https://github.com/haroutboujakjian/Vuesalize).
- [Django Rest Framework](https://www.django-rest-framework.org/) is added to the backend to create an API endpoint for retrieving the data for the dashboard. This can be used as a starting point for eventually migrating the entire app to a separate SPA and REST API.
- Almost all statistics/aggregates for the dashboard are calculated in the database rather than python or client side.
- Responsive design that allows dashboard to function on smaller devices (e.g. ipad, iphone, etc.)


ajbarnes:
I ran through this and caught some of the things that the CI tests were complaining about + bumped the things we were already upgrading to the most recent versions. I did get stuck however on two tests in particular:
- `va_explorer/va_data_management/tests/test_views.py::test_index_redacted`
- `va_explorer/home/tests/test_views.py::test_trends`

These weren't the focus of this PR but it seems like the change to `users.verbal_autopsies()` has broken something in a bit of an unintended chain.. Let's sync when you are available again.

ajbarnes (approved):
Having pulled this to fix the CI complaints, I'm observing expected behavior now and the code review looks good. I've recorded some small tangential things during my review, but think they're best  opened as separate issues. LGTM, and look forward to feedback from the team in Zambia.

PR #253, opened 2022-11-15
Improve dashboard usability

haroutboujakjian:
This PR adds the followings fixes to the dashboard:

- Add a scrollbar to the COD chart to allow for more room between bars
- Display COD group codes in dropdown based on env var
- Fix issue with null value in district/province dropdown in Geographic Distribution panel


ajbarnes (approved):
Pulled this myself to investigate and I think it looks good. LGTM

PR #254, opened 2022-11-28
Add filters for age and sex, and performance tweaks

haroutboujakjian:
This PR focuses on the following items:

1. Added filters for age and sex in the dashboard
2. Added an index for death date to speed up filtering
3. Saved `get_va_summary_stats` to cache, using DB as backup if not in cache
4. Added loading indicator to dashboard so user is aware the data is being fetched


ajbarnes (approved):
Pulled locally and tested/ ran tests. Looks great and I believe is notably faster. I added a small tweak to age filtering to account for more va spec props. LGTM

PR #256, opened 2022-12-23
Update/modernize documentation

ajbarnes:


ajbarnes (commented):
Previous reviewer left project so I'm resorting to a self-review in this limited case. Have run tests and linting locally, run both the documentation and app locally, and built the pdf locally. All seem to be working as expected.

All documentation goals accomplished here except for correcting internal references and finishing the glossary. I will open a separate issue for those so this doesn't become a blocker. Squashed commit should acknowledge version bumps for ruff that result in linting changes + small bug fixes.

PR #273, opened 2023-03-07
270 fix deceased filtering with spaces

### LINKED ISSUE #270: Search/filter on data management page does not handle spaces
**Describe the bug**
When searching for any field via the search form on the data management page, the search is successful for one word, but unsuccessful for multiple words (ex. firstname lastname)

**To Reproduce**
1. Ensure VAs are loaded that contain multi-word phrases for at least one of the searchable fields on the data management page
2. Attempt to search for the first word in the target phrase and confirm hits
3. Add a space plus the second to the target phrase and confirm no hits

**Expected behavior**
If my search for is a multi-word phrase such as "firstname lastname" on the name of the interviewer I expect have my search return all VAs with "firstname lastname" as the interviewer. Likewise for any other searchable/filterable field

kji-mitre:
Resolves #270  

Modifies the search/filter behavior for the deceased field on the data management page to handle search terms containing spaces:
- Performs `fuzzy_token_set`  matching against the combined `Id10017` and `Id10018` fields joined by a space

### DIFF
--- a/va_explorer/va_data_management/filters.py
+++ b/va_explorer/va_data_management/filters.py
@@ -1,8 +1,8 @@
 from django import forms
 from django.core.exceptions import ValidationError
-from django.db.models import Q
 from django.forms import Select, TextInput
 from django_filters import BooleanFilter, CharFilter, DateFilter, FilterSet
+from fuzzywuzzy import fuzz
 
 from .models import VerbalAutopsy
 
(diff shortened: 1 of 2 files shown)


ajbarnes (approved):
Pulled and confirmed for myself and I think it looks good/ fixes the bug. I tweaked the threshold a bit to see how the results changed and agree 75% seems to work well/ be a balanced target.

PR #276, opened 2023-03-21
Replace refs to submissiondate with Id10012 (Interview date)

### LINKED ISSUE #263: Change date used to report VAs
When I view VAs as any type of user I expect to see VAs dated via the date of interview field (`Id10012`).

**What is the actual state?**
When I view VAs as any type of user I see VAs dated via the `submissiondate` field

**Relevant context**
endusers report that `submissiondate` is unreliable as it is
1. generated automatically via VA collection software and sometimes that process fails
2. can be inaccurate due to batch submission of multiple VAs at the same time that were conducted on days other than `submissiondate` 

and that date of interview `Id10012` is required to be filled out before submission and reflects the actual date of the VA being filled out. Due to 1, we also observe 'dk' as the `submissiondate` sometimes and this issue would fix that.

ajbarnes:
This PR replaces all functional references to `va.submissiondate` with `va.Id10012` (the interview date instead) for reasons mentioned in #263.

- Updates homepage charts & dashboard to use Id10012 and changes language from "submitted" to "interviewed" and similar tweaks
- Updates data update boxes on home & dashboard to use `last_interview` instead of `last_submission`
- Updates supervision page to calculate interviews based off interview date instead of submission date
- Updates other modules like data management, VA partial table, cleanup, etc to use the correct field
- Updates management commands that create fake dates for testing/demo to set Interview date instead of submission
- Fixes 2 small bugs found while implementing, 1 preventing correct sorting on the VA table partial, the other preventing local runs of `run_coding_algorithms`

Closes #263

### DIFF
--- a/va_explorer/static/js/dashboard.js
+++ b/va_explorer/static/js/dashboard.js
@@ -207,7 +207,6 @@ const dashboard = new Vue({
             this.demographics.forEach(d => {
                 delete d.order;
             });
-            console.log(this.demographics)
 
             // display warning message if small sample size (allow user to stop these dialogs)
             if (!this.suppressWarning &&
--- a/va_explorer/static/js/dashboard.js
+++ b/va_explorer/static/js/dashboard.js
@@ -24,7 +24,7 @@ const dashboard = new Vue({
             uncoded_vas: 0,
             update_stats: {
                 last_update: 0,
-                last_submission: 0,
+                last_interview: 0,
             },
 
             // chart sizes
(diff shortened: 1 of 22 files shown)


va_explorer/static/js/dashboard.js
```diff
@@ -201,6 +201,7 @@ const dashboard = new Vue({
             this.demographics.forEach(d => {
                 delete d.order;
             });
+            console.log(this.demographics)
```
ajbarnes, line 204: need to remove

kji-mitre (approved):
Looks good to me!

PR #277, opened 2023-03-25
Fix various small bugs in prep for PoP end

ajbarnes:
- **Fixes `run_coding_algorithms` bug where command was reporting the length of the string keys instead of the length of the array values**
- Restricts use of Run Coding Algorithms button to production environments only since it only works in a docker-compose deployment context. Use `./manage.py run_coding_algorithms` instead for local coding.
- Refactors app to use fontawesome, trading very size-efficient fontello for simpler to understand/use fontawesome
  - Uses fontawesome to fix broken/missing icons in various places throughout the app
- Fixes _generic_table partial to use 0 as default instead of "Unknown" if value is number because 0 was triggering "Unknown" 
- Fixes VA validator to use missing `ageInYears2` as cause for warning instead of `ageInYears`
- **Patches import to support the custom field SubmitterName and use it to replace Id10010 if missing**


kji-mitre (approved):
Looks good to me!

PR #282, opened 2023-05-03
Support initial KoboToolbox integration

ajbarnes:
- [x] Implements import_from_kobo management command
- [x] Adds some basic tests for the feature
- [x] Updates docs with new integration info

- Adds KOBO_HOST, KOBO_API_TOKEN, & KOBO_ASSET_ID env variables for integration configuration
- Adds import_from_kobo capable of importing kobo VA data into VA Explorer, connecting it up with existing processes/validations implemented for ODK integration.
- Adds get_kobo_api_token as optional util method for obtaining a kobo api token given kobo username and password

Leaves, as a future exercise, implementing an endpoint to support Kobo's REST services feature, which is different from this feature in that the REST services feature requires VA Explorer to receive an automatic incoming push of data (whenever someone fills out a single Verbal Autopsy) from Kobo instead of a pull of all Verbal Autopsies like we do here.


va_explorer/va_data_management/utils/kobo.py
```diff
@@ -0,0 +1,53 @@
+import os
+
+import pandas as pd
+import requests
+from requests.auth import HTTPBasicAuth
+
+KOBO_HOST = os.environ.get("KOBO_HOST", "http://127.0.0.1:5003")
+# Don't verify localhost (self-signed cert)
+SSL_VERIFY = os.environ.get(
+    "KOBO_SSL_VERIFY", not KOBO_HOST.startswith("http://localhost")
```
mboyas-mitre, line 10: IF KOBO_SSL_VERIFY is set in .env, it appears to read in here as a string type "True" or "False" instead of Boolean, which then causes everything to fail.  NOT having it set in .env works just fine.

Proposal for consideration:  at line 12, add: SSL_VERIFY = bool(SSL_VERIFY)

NOTE: I *think* this issue may also appear in odk.py

ghost (changes requested):
Aside from the one bug if ssl verify is set, seems to work well. Was able to import from the test kobo server, run algorithms, and things appear in VAE just fine.

ghost (approved):

PR #284, opened 2023-06-02
Introduce logic for removal to the import process

ajbarnes:
Motivated by the need to support edits that take place on kobo and updating the relevant records in VAE on subsequent pulls, the shared `loading.py` has been augmented to handle removal logic.

It mainly accomplishes this by now *requiring* an `instancename` value be associated with every VA along with the `instanceid` that has always been required. This is necessary because when a record is edited in kobo, its uuid will change. Because of this VAE would have imported it as a new VA and left the old instance in place. `instancename` as something that remains the same between edits was chosen because it is present in both ODK and Kobo so shouldn't break compatability.

Additionally, removal will now be supported for kobo specifically because its records have the `_validation_status` column. This PR introduces logic to now check for that and remove VAs that are marked invalid.

### Overall:
- logic to remove invalid VAs via kobo-specific `_validation_status` field implemented
- logic to remove 'outdated' VAs implemented using `instancename` field implemented
- tests and other import integrations updated to handle the newly tracked `instancename` field


mboyas-mitre (approved):
Looks good to me - the previously noted issues tested just fine.

PR #285, opened 2023-07-28
Support Kobo Pagination & make instancename workaround more robust

ajbarnes:
This PR
- Adds logic to `import_from_kobo.py` to support paginated downloads. Necessary if data in kobo is larger than max page size of 30k
- Adds logic to `kobo.py` to support an external docker network. Necessary if VA Explorer and the service one wished to integrate with (Kobo in this case) are both run docker-compose and both on the same server host.
  - Adds documentation explaining how to configure new network settings
- Adds logic to `loading.py` to handle VAs that may be missing `instancename` by implementing a workaround that requires `Id10017` (first name of deceased), `Id10018` (last name of deceased), and `Id10012` (date of interview), to construct one manually as a workaround. VAs missing still missing these minimal fields are considered invalid for the purposes of import.
  - Updates test data to handle these new field expectations

### DIFF
--- a/docs/usage/integrations.md
+++ b/docs/usage/integrations.md
@@ -12,6 +12,56 @@ integrations as well as details on how to enable them in VA Explorer if needed.
   - pyCrossVA
   - InterVA5
 
+## Networking
+
+When integrating services with VA Explorer, operators should be aware of additional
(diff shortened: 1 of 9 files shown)


mboyas-mitre (approved):
Tested with test instance, confirmed working

PR #286, opened 2023-10-05
Remove pilot legacy event logging

ajbarnes:
Since event_logger is causing issues with new installations and we don't have the bandwidth to make logging more robust at the moment, opt to remove legacy event_logging that was originally implemented for pilot purposes.

This PR:
- removes event_logger
- removes ingest_logger
- removes calls to either of the above
- removes va_logs django sub application
- removes va_explorer_logs at the root level and things that target it like docker volumes


mboyas-mitre (approved):

mboyas-mitre (changes requested):
Need to also remove the line `python /app/manage.py initialize_log_directory` in `./compose/django/start` to allow deploying via Docker.

mboyas-mitre (approved):

PR #292, opened 2025-06-04
update copyright year 2023 -> 2025

ghost:


ajbarnes (approved):
Consider shifting supported versions to 3.11, 3.12, and 3.13.

3.8 is already eol so black 25.1.0 used by pre-commit no longer works for that CI check. 3.9 / 3.10 eol are imminent. 3.13 is the currently suggested latest.

For now, LGTM

